PR #25, opened 2026-05-20
fix(startup-banner): correct AGENTS/SKILLS/EXTENSIONS stats and add info dashboard

aporcelli:
## Problem

The startup banner HUD displays inaccurate information:

- **AGENTS** label counts skill-registered commands from `pi.getCommands()`, not actual SDD agent files
- **EXTENSIONS** reads `settings.extensions[]` which is always empty; real extensions are declared in each installed package's `pi.extensions` manifest
- No distinction between SDD phase agents and on-demand skills

### AGENTS → AGENTS + SKILLS (split)

| Before | After |
|--------|-------|
| `AGENTS: N loaded` (skill commands) | `AGENTS: N phases` (SDD agent files) |
| | `SKILLS: N loaded` (skill commands) |

- **AGENTS** counts files in `~/.pi/agent/agents/sdd-*.md`
- **SKILLS** keeps the original `pi.getCommands().filter(source==="skill")` count

### EXTENSIONS: now counts real extension files

Scans each installed package's `package.json` and sums `pi.extensions` entries, instead of reading the always-empty `settings.extensions[]`.

### Dashboard layout (wide mode)

```
PATH:       /home/user/project        EXTENSIONS: N active
GIT:        On branch main            PLUGINS:    N package(s)
MCP:        N server(s)               SKILLS:     N loaded
AGENTS:     N phases                  TOOLS:      N custom
VER:        v0.75.4
```

### Technical changes

- Added `readdir` import for counting SDD agent files
- Added `countSddAgents()` — reads `~/.pi/agent/agents/` for `sdd-*.md` files
- Fixed `countPackageExtensions()` — resolves each npm package's `package.json`, reads `pi.extensions`, sums counts. Fixed early `return null` bug that aborted the entire function on git packages
- Removed broken `settings.extensions[]` lookup
- Reordered dashboard rows: Left: PATH, GIT, MCP, AGENTS, VER | Right: EXTENSIONS, PLUGINS, SKILLS, TOOLS

## Changes

| File | Change |
|------|--------|
| `extensions/startup-banner.ts` | Enhanced HUD: split AGENTS/SKILLS, fixed EXTENSIONS count, reordered layout |

## Test Plan

- [x] Verified AGENTS shows correct count of `sdd-*.md` files in agent dir
- [x] Verified SKILLS count matches `pi.getCommands()` output
- [x] Verified EXTENSIONS matches sum of `pi.extensions` across installed packages
- [x] Verified PLUGINS count matches `settings.packages` length
- [x] Verified layout renders correctly in both wide and narrow terminal widths

### DIFF
--- a/extensions/startup-banner.ts
+++ b/extensions/startup-banner.ts
@@ -369,12 +369,38 @@ class LayoutBuilder {
   }
 }
 
+const FULL_INTRO_MIN_ROWS = 30;
+const FULL_INTRO_MIN_COLS = 80;
+const MINIMAL_INTRO_MIN_ROWS = 20;
+const MINIMAL_INTRO_MIN_COLS = 40;
+const RESIZE_DEBOUNCE_MS = 150;
+const RESIZE_GRACE_PERIOD_MS = 300;
+
+type IntroMode = "full" | "minimal" | "skip";
+
+function pickIntroMode(rows: number, cols: number): IntroMode {
+  if (rows >= FULL_INTRO_MIN_ROWS && cols >= FULL_INTRO_MIN_COLS) return "full";
+  if (rows >= MINIMAL_INTRO_MIN_ROWS && cols >= MINIMAL_INTRO_MIN_COLS) return "minimal";
+  return "skip";
+}
+
+function currentIntroMode(): IntroMode {
+  const rows = process.stdout.rows ?? 0;
+  const cols = process.stdout.columns ?? 0;
+  return pickIntroMode(rows, cols);
+}
+
 export default function (pi: ExtensionAPI) {
   pi.on("session_start", async (_event, ctx) => {
     if (!ctx.hasUI) return;
 
-    // En modo de prueba con `-e <archivo.ts>` sí queremos mostrar la intro,
-    // así que no filtramos por argv en esta versión.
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman (changes requested):
Thanks for the cleanup here, the direction for splitting AGENTS/SKILLS and counting real extension manifests makes sense.

I’m requesting changes because this PR also removes the existing startup gating around CLI commands and small/unsupported terminal modes. That changes behavior outside the stats/dashboard fix: the banner can clear/render during flows like `pi update`, `pi install`, scripted runs, or constrained terminals where current main intentionally skips or downgrades the intro.

Can you restore the CLI/TTY/terminal-size/resize guard semantics from current main while keeping the stats/dashboard changes? Also please remove or finish the new sparkle path, since `renderStarCell()` currently always returns `null` and leaves the sparkle variables effectively dead.

Alan-TheGentleman:
Thanks for the dashboard cleanup. I kept the existing CLI/TTY/terminal-size guard semantics from main and landed the safe stats part directly in 836a26d6: AGENTS now counts SDD phase agents, SKILLS keeps skill-command count, and EXTENSIONS counts package manifest `pi.extensions`. Closing this PR as superseded by that main commit rather than merging the stale branch with the removed startup gates.

PR #28, opened 2026-05-21
feat(startup): add /intro toggle and startupAnimation shortcuts

aporcelli:
## Summary
- add `/intro on|off|disable` command
- `on`: animated intro + HUD
- `off`: static intro + HUD
- `disable`: HUD-only (no rose/logo intro rendering)
- `F1` now cycles intro modes in order: `on -> off -> disable -> on`
- handle missing/corrupt `~/.pi/agent/settings.json` safely with resilient read/write helpers

## Dependency
- Depends on #25
- This branch was created from `fix/startup-banner-accuracy` (PR #25 head)
- Please merge #25 first, then merge this PR

## Notes
- `ctrl+n` intentionally avoided due built-in keybinding conflict.

### DIFF
--- a/extensions/startup-banner.ts
+++ b/extensions/startup-banner.ts
@@ -393,61 +393,89 @@ function currentIntroMode(): IntroMode {
 export default function (pi: ExtensionAPI) {
   const settingsPath = join(os.homedir(), ".pi", "agent", "settings.json");
 
-  const readStartupAnimation = async (): Promise<boolean> => {
+  const readSettings = async (): Promise<Record<string, unknown>> => {
     try {
       const raw = await readFile(settingsPath, "utf8");
-      const cfg = JSON.parse(raw);
-      return cfg.startupAnimation !== false;
+      const parsed = JSON.parse(raw);
+      return parsed && typeof parsed === "object" ? parsed : {};
     } catch {
-      return true;
+      return {};
     }
   };
 
-  const writeStartupAnimation = async (enabled: boolean): Promise<void> => {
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman (changes requested):
Thanks, this is useful, but I’m going to hold this one until #25 is fixed and merged.

This branch is stacked on #25 and currently inherits the startup gating regression from that PR. Once #25 restores the CLI/TTY/terminal-size guards, please rebase or refresh this branch so this review can focus only on the `/intro` toggle behavior.

Non-blocking follow-up for the refresh: the wording should probably say this toggles the startup animation, not the whole intro, because `off` still renders the static banner. It would also be good to handle a missing or corrupt `~/.pi/agent/settings.json` gracefully.

Alan-TheGentleman:
I reviewed this again after landing the safe dashboard-stat fixes from #25 in 836a26d6. I am not merging this branch as-is because it is still stacked on the stale #25 branch and inherits the startup-gating regression noted in review. The desired direction is still useful: HUD-only/disabled intro mode is the right shape. Please refresh this PR on current main so the diff contains only the `/intro` mode behavior and preserves the CLI/TTY/terminal-size guards.

Alan-TheGentleman:
Actualización: la configuración del banner ya quedó implementada en main en 83299a8b, y el panel puede quedar visible aunque la rosa y el logo estén ocultos. Este PR sigue stackeado sobre una rama vieja con regresión de guards, así que lo cierro como reemplazado por la implementación actual en main.

PR #127, opened 2026-07-12
fix(review): accept absolute Windows Git common directories

MarsSall:
Closes #115

## Summary

- Accept platform-native absolute Git common-directory paths using Node's `path.isAbsolute`.
- Add a Windows regression covering drive-letter Git common directories.

## Type

- [x] Bug fix

## Changes

| File | Change |
|------|--------|
| `lib/review-repository.ts` | Replace the POSIX-only leading-slash check with platform-aware absolute-path validation. |
| `tests/review-repository.test.ts` | Verify repository authority resolution with a Windows drive-letter common directory. |

## Test plan

- [x] Focused Windows regression passes.
- [x] Staged diff validation passes.
- [x] Bounded reliability review approved.

## Checklist

- [x] Linked issue #115.
- [x] Conventional commit used.
- [x] No unrelated fsync portability changes included.
- [x] No AI attribution or co-author trailers included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository validation to correctly recognize absolute paths across operating systems, including Windows drive-letter paths.
  * Repository identity information is now resolved correctly for Windows-based Git configurations.

Issue #115, opened 2026-07-12
fix(review): accept absolute Windows Git common directories


MarsSall:
## Bug description

The bounded review controller rejects valid Windows repositories because `probeLiveRepositoryV1()` checks whether Git's common directory starts with `/`.

Git correctly returns an absolute Windows path such as:

```text
C:/Users/example/project/.git
```

However, `lib/review-repository.ts` currently validates it with:

```ts
if (!commonDirectory.startsWith("/")) {
  throw new ReviewRepositoryError("Git common directory is not absolute");
}
```

This Unix-only check rejects drive-letter paths and prevents `gentle_review validate` from authorizing lifecycle commands on Windows.

## Steps to reproduce

1. Install `gentle-pi` 1.0.2 on Windows.
2. Open a Git repository through Pi.
3. Start or reuse an approved bounded review transaction.
4. Run pre-commit validation through `gentle_review validate`.
5. Observe `Git common directory is not absolute`.
6. Run `git rev-parse --path-format=absolute --git-common-dir` and observe that Git returned a valid drive-letter absolute path.

## Expected behavior

The review controller should accept platform-native absolute paths on Windows, macOS, and Linux.

## Actual behavior

Valid Windows paths are rejected because they do not start with `/`.

## Proposed fix

Use Node's platform-aware path validation:

```ts
import { isAbsolute } from "node:path";

if (!isAbsolute(commonDirectory)) {
  throw new ReviewRepositoryError("Git common directory is not absolute");
}
```

Add regression coverage for a Windows drive-letter path.

## Environment

- OS: Windows
- `gentle-pi`: 1.0.2
- Pi Coding Agent: 0.80.6
- Git output example: `C:/Users/mjsal/Desarrollos IA/Modificacion de Terminales/engram/.git`

## Duplicate check

Searched existing issues and pull requests for Windows, `git-common-dir`, absolute path, and review controller failures; no matching report was found.

rafaga2469:
Still reproducing on gentle-pi 1.0.6 + Pi 0.80.10, Windows 11. `gentle_review validate` fails with `Git common directory is not absolute` even though `git rev-parse --git-common-dir` returns a valid `D:/workspace/sgi-timetracker/.git`.

[...]

PR #131, opened 2026-07-12
fix(review): make persistence durability portable on Windows

### LINKED ISSUE #128: fix(review): make persistence durability portable on Windows
## Pre-flight checks

- [x] I searched existing issues and pull requests for Windows, `fsync`, `EPERM`, directory durability, and review persistence; no duplicate was found.
- [x] I understand the change requires focused review before merge.

## Bug description

The bounded-review persistence layer assumes POSIX `fsync` behavior. On Windows, Node rejects `fsyncSync()` for read-only file descriptors and directory descriptors with `EPERM`, preventing review authority creation, mutation, validation, reset, bundle, and recovery workflows.

This becomes visible after valid Windows Git common-directory paths are accepted by #115.

## Steps to reproduce

1. Run gentle-pi on Windows in a Git repository without previously initialized review authority.
2. Call `gentle_review inspect` or start an ordinary bounded review.
3. Allow the controller to persist `IDENTITY`, lock metadata, graph objects, compact state, checkpoints, or reset state.
4. Observe Windows reject an `fsyncSync()` call on a descriptor opened with `"r"` or on a directory descriptor.

Minimal Node reproduction:

```js
const fs = require("node:fs");
const fd = fs.openSync("file.txt", "r");
fs.fsyncSync(fd); // EPERM on Windows
```

## Expected behavior

Review persistence should work on Windows while preserving:

- writable-file durability;
- POSIX directory durability where supported;
- same-directory atomic replacement for mutable state;
- hard-link/no-replace semantics for immutable publication;
- lock ownership, quorum, retry, and fail-closed guarantees.

## Actual behavior

The controller fails with:

```text
EPERM: operation not permitted, fsync
```

The failure appears sequentially across repository identity, mutation locks, graph object storage, compact storage, checkpoints, transactions, reset state, bundles, and supersession records.

## Proposed correction

- Flush files through writable descriptors or retain the descriptor used for writing.
- Preserve directory `fsync` on POSIX and explicitly skip the unsupported directory operation on Windows.
- Keep native same-directory atomic rename for mutable replacement.
- Preserve hard-link/no-replace publication for immutable records and bundles.
- Add focused Windows lifecycle and repeated-publication regressions.

## Operating system

Windows

## Agent / client

Pi Coding Agent with gentle-pi

## Shell

PowerShell / Git Bash

## Verification evidence

Focused Windows suites passed after the local correction:

- repository identity regression;
- mutation lock suite: 8/8;
- graph object-store suite: 9/9;
- combined targeted persistence suites: 83 passed with one unrelated authorization assertion failure;
- reset, supersession, and bundle suites: 43/43.

## Scope note

This issue is intentionally separate from #115. That issue fixes platform-aware Git absolute-path validation; this issue covers the subsequent Windows persistence failures.

MarsSall:
Closes #128

## Summary

- Make review persistence file flushing compatible with Windows by using writable descriptors.
- Preserve directory durability on POSIX while explicitly skipping unsupported directory `fsync` on Windows.
- Preserve mutable atomic replacement, immutable no-replace publication, lock ownership, and retry/conflict semantics.

## Dependency

This PR is independent at the code level but its end-to-end Windows repository flow becomes reachable after #127 fixes Windows Git common-directory validation.

```text
#127 Windows Git absolute paths
  └── #128 Windows persistence durability 📍
```

## Type

- [x] Bug fix

## Changes

| Area | Change |
|------|--------|
| Repository identity, locks, graph and compact stores | Use writable file descriptors and portable directory durability. |
| Checkpoints and legacy transactions | Preserve atomic publication while avoiding unsupported Windows directory flushes. |
| Reset, bundles, and supersession | Keep mutable replacement and immutable no-replace guarantees across platforms. |
| Windows supersession fixtures | Normalize path separators and avoid asserting POSIX-only permission modes. |

## Test plan

- [x] Direct persistence suites: 17/20; all three failures were the separate #115/#127 path prerequisite.
- [x] Temporary #127 prerequisite verification observed no Windows `EPERM: fsync` failures.
- [x] Supersession suite after Windows fixture corrections: 17/17.
- [x] Lock no-replace, reset, checkpoint, bundle, and immutable publication behavior passed.
- [x] `git diff --check` passed.
- [x] Bounded reliability review approved with no findings.

## Known unrelated failure

The existing annotated-tag PUSH CREATE authorization assertion remains outside this persistence change.

## Checklist

- [x] Linked issue #128.
- [x] Conventional commit used.
- [x] Diff is focused: 68 changed lines across 10 files.
- [x] Issue #115/#127 implementation is not duplicated.
- [x] No temporary fork-head gate patch included.
- [x] No AI attribution or co-author trailers included.

### DIFF
--- a/lib/review-authority-supersession.ts
+++ b/lib/review-authority-supersession.ts
@@ -594,6 +594,7 @@ export interface SupersessionWriteOptionsV1 {
 }
 
 function fsyncDirectoryV1(directory: string): void {
+	if (process.platform === "win32") return;
 	const handle = openSync(directory, "r");
 	try { fsyncSync(handle); } finally { closeSync(handle); }
 }
@@ -624,7 +625,7 @@ export function writeSupersessionRecordV1(directory: string, envelope: Supersess
 	try {
 		writeFileSync(temporary, bytes, { flag: "wx", mode: 0o600 });
 		options.faultInjector?.("before-temp-fsync");
-		const file = openSync(temporary, "r");
+		const file = openSync(temporary, "r+");
 		try { fsyncSync(file); } finally { closeSync(file); }
 		options.faultInjector?.("before-link");
 		try { linkSync(temporary, path); } catch (error) {
@@ -844,7 +845,7 @@ export class SupersessionStoreV1 {
 		const path = this.recoveryMarkerPath(changeName, true);
 		if (!this.isRecoveryRequiredMarker(path)) {
 			writeFileSync(path, canonicalJsonV1({ schema: "gentle-ai.recovery-required/v1", change_name: changeName }), { flag: "wx", mode: 0o600 });
-			const file = openSync(path, "r");
+			const file = openSync(path, "r+");
 			try { fsyncSync(file); } finally { closeSync(file); }
 		}
 		fsyncDirectoryV1(this.recoveryMarkerDirectory(false));
--- a/lib/review-bundle.ts
+++ b/lib/review-bundle.ts
@@ -137,8 +137,8 @@ function closure(store: ReviewGraphObjectStoreV1, roots: readonly ReviewBundleRo
 	return events;
 }
 
-function fsyncFile(path: string): void { const fd = openSync(path, "r"); try { fsyncSync(fd); } finally { closeSync(fd); } }
-function fsyncDirectory(path: string): void { if (!statSync(path).isDirectory()) throw new ReviewBundleError("Bundle destination parent is invalid"); const fd = openSync(path, "r"); try { fsyncSync(fd); } finally { closeSync(fd); } }
+function fsyncFile(path: string): void { const fd = openSync(path, "r+"); try { fsyncSync(fd); } finally { closeSync(fd); } }
+function fsyncDirectory(path: string): void { if (!statSync(path).isDirectory()) throw new ReviewBundleError("Bundle destination parent is invalid"); if (process.platform === "win32") return; const fd = openSync(path, "r"); try { fsyncSync(fd); } finally { closeSync(fd); } }
 function writeCheckpoint(checkpoints: ReviewCheckpointStoreV1, input: Parameters<ReviewCheckpointStoreV1["write"]>[0]): void {
 	let sequence = 0;
 	try { sequence = checkpoints.read(input.operation_id, input).checkpoint_sequence + 1; } catch (error) { if (!(error instanceof Error) || !/missing or malformed/.test(error.message)) throw error; }
--- a/lib/review-checkpoint.ts
+++ b/lib/review-checkpoint.ts
@@ -122,6 +122,6 @@ export class ReviewCheckpointStoreV1 {
 		return join(this.root, `${operationId}.json`);
 	}
 
-	private fsyncFile(path: string): void { const descriptor = openSync(path, "r"); try { fsyncSync(descriptor); } finally { closeSync(descriptor); } }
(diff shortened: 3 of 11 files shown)


Ividni:
Verified this persistence patch on top of `v1.0.6` plus PR #127, with raw and causal results kept separate.

Fresh Windows evidence:

- PR #127-only baseline: 17/84 tests passed, 67 failed, primarily at persistence setup.
- PR #127 + PR #131 raw result: 77/84 passed.
- Candidate-causal durability selection: 77/77 passed.
- Six residual failures are Windows symlink-fixture `EPERM` errors before the product assertions.
- The remaining annotated-tag expected-deny/actual-allow failure is baseline-identical and matches the PR's documented unrelated failure.
- The exact PR #131 diff applies cleanly to `v1.0.6`.

I am not claiming the raw suite is green. The evidence shows the fsync/persistence changes resolve the intended Windows durability failures without broadening production behavior. The symlink fixtures and annotated-tag assertion should remain separate follow-ups.

Alan-TheGentleman (changes requested):
Hey MarsSall! Thanks for this — the overall pattern is exactly right: `"r"` → `"r+"` for file fsync (Windows needs write access for FlushFileBuffers) and skipping directory fsync on win32 are both the correct portable moves, and the test guards are reasonable. One blocking issue before this can merge:

## The win32 branch breaks the no-replace guarantee for files

In `lib/review-lock.ts`, the new win32 branch runs before the `isDirectory` check and uses plain `renameSync(source, destination)`:

```ts
if (process.platform === "win32") {
    try {
        renameSync(source, destination);
    } catch (error) { ... }
    return;
}
```

Node's `fs.rename` on Windows goes through libuv's `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`, so if the destination **file** already exists it is silently replaced. The POSIX path uses `linkSync` precisely because `link(2)` fails with `EEXIST` — that failure IS the no-replace contract this function's name promises. With this branch, a contended lock on Windows gets silently stolen instead of surfacing contention, which breaks the mutual-exclusion guarantee the whole authority store leans on.

## Solution

NTFS supports hard links, and `CreateHardLinkW` (what `linkSync` uses on Windows) fails if the destination exists — same EEXIST semantics as POSIX. So files don't need a win32 special case at all; only directories do (they can't be hard-linked, and Windows `rename` onto an existing directory fails anyway, which preserves no-replace):

```ts
if (!isDirectory) {
    // linkSync + unlinkSync works on win32/NTFS too and keeps EEXIST no-replace semantics
    try { linkSync(source, destination); } catch ...
    ...
} else if (process.platform === "win32") {
    // Windows rename refuses to replace an existing directory, so this stays no-replace
    try { renameSync(source, destination); } catch ...
    return;
}
```

Ideally with a regression test asserting that a pre-existing destination file makes the move throw (that test is platform-independent — the EEXIST behavior via `linkSync` holds on POSIX and NTFS alike).

---

Everything else looks good to go — fix the lock branch and this merges. It's currently a release blocker for v1.1.0 Windows support, so if you're short on time let us know and we can push the fixup.

Alan-TheGentleman (approved):
Fixup applied directly on the branch (maintainer edit): lock files now keep the linkSync EEXIST no-replace semantics on every platform (NTFS included), with the win32 rename special case retained only for directories, plus dual-platform regression tests proving an occupied destination file/directory is never silently replaced. With that, the whole change is solid — thanks again MarsSall!

PR #145, opened 2026-07-14
fix: normalize win32 platform to windows for asset lookup

### LINKED ISSUE #141: fix: resolveGentleAiReleaseAsset fails on Windows (win32 vs windows platform mismatch)
## Bug

`resolveGentleAiReleaseAsset()` in `scripts/gentle-ai-installer.mjs` fails on Windows with:

```
unsupported Gentle AI platform/architecture: win32/x64; supported pairs are darwin, linux, or windows with x64 or arm64
```

## Root cause

Node's `process.platform` returns `"win32"` on Windows, but the release asset map uses `"windows"` as the platform key:

```js
// L37 — key is "${platform}/${arch}" → "win32/amd64"
const key = `${platform}/${upstreamArchitecture(architecture)}`;

// But the map has "windows/amd64", not "win32/amd64"
GENTLE_AI_RELEASE_ASSETS = {
  "windows/amd64": asset(...),  // ← "windows"
  "windows/arm64": asset(...),  // ← "windows"
};
```

The `upstreamArchitecture()` function correctly maps `x64 → amd64`, but there's no equivalent `upstreamPlatform()` to map `win32 → windows`.

## Fix

Add a platform mapping function and use it when building the key:

```js
function upstreamPlatform(platform) {
    return platform === "win32" ? "windows" : platform;
}

export function resolveGentleAiReleaseAsset(...) {
    const key = `${upstreamPlatform(platform)}/${upstreamArchitecture(architecture)}`;
    // ...
}
```

Or alternatively, change the asset map keys from `"windows"` to `"win32"` and adjust the trusted extractor check accordingly.

sombraxxl:
## Problem

\process.platform\ returns \win32\ on Windows, but the asset map in \GENTLE_AI_RELEASE_ASSETS\ uses \windows/amd64\ and \windows/arm64\ keys. This causes the installer to fail with:

\\\
unsupported Gentle AI platform/architecture: win32/x64
\\\

Even though the Windows binaries exist in the v2.1.4 release.

## Fix

Added \upstreamPlatform()\ function that normalizes \win32\ → \windows\, mirroring the existing \upstreamArchitecture()\ pattern for \x64\ → \^Gmd64\.

## Test

Added test \win32 platform normalizes to windows for asset lookup\ that verifies both \win32/x64\ and \win32/arm64\ resolve correctly.

## Verification

Tested on Windows 11 x64 — the binary downloaded and installed successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows installer asset resolution so the correct release files are selected for both x64 and arm64 systems.

Closes #141

### DIFF
--- a/scripts/gentle-ai-installer.mjs
+++ b/scripts/gentle-ai-installer.mjs
@@ -42,8 +42,12 @@ function upstreamArchitecture(architecture) {
 	return architecture === "x64" ? "amd64" : architecture;
 }
 
+function upstreamPlatform(platform) {
+	return platform === "win32" ? "windows" : platform;
+}
+
 export function resolveGentleAiReleaseAsset(platform = process.platform, architecture = process.arch, releaseAssets = GENTLE_AI_RELEASE_ASSETS) {
-	const key = `${platform}/${upstreamArchitecture(architecture)}`;
+	const key = `${upstreamPlatform(platform)}/${upstreamArchitecture(architecture)}`;
 	const resolved = releaseAssets[key];
 	if (!resolved) throw new Error(`unsupported Gentle AI platform/architecture: ${platform}/${architecture}; supported pairs are darwin, linux, or windows with x64 or arm64`);
 	return resolved;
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (approved):
Verified the Windows platform normalization against the exact PR head. The focused installer suite and the full test suite pass, unsupported platforms remain fail-closed, and the change is correctly scoped. Approved.

PR #151, opened 2026-07-15
 fix(review): preserva diagnósticos ante módulos duplicados

jacargentina:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native error handling when operating across multiple module instances, ensuring consistent classification and “next action” behavior.
  * Blocked responses now preserve valid diagnostic details while redacting sensitive standard-error output.
  * Malformed or unexpected diagnostics are filtered out so they no longer appear in response details.
* **Tests**
  * Added coverage for ambiguous native start failures, including cases with valid diagnostics and cases with malformed diagnostics.

### DIFF
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -141,8 +141,10 @@ import {
 	NATIVE_REVIEW_AUTHORITY_STATUS,
 	NATIVE_REVIEW_ERROR_CODE,
 	NATIVE_REVIEW_LOCK_STATUS,
+	sanitizeForeignNativeReviewDiagnostics,
 	type NativeReviewCli,
 	type NativeFinalizeResult,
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman (approved):
Maintainer fixup applied on the branch (edits enabled): rebased onto current main resolving the extensions conflicts while preserving both your semantics (foreign-instance error classification with the operation-match guard on diagnostics) and today's landed work; regenerated the runtime mirror your change needed (lib/native-review-cli.ts is machine-mirrored into runtime/ and CI's drift check would have failed); updated the two test envelopes to the post-#186 ambiguous-mutation contract (the substantive foreign-diagnostics assertions are intact). The sanitizer's exactObject field lists were verified against the current NativeReviewProcessDiagnostics shape — no drift. Solid fix for a subtle problem, thanks Javier!

Issue #215, opened 2026-07-21
bug(review): native snapshots drop gitlinks and create divergent review authority


jacargentina:
## Pre-flight checks

- [x] I searched open and closed issues for `submodule`, `gitlink`, `review authority`, `workspace projection`, and Markdown-only commits.
- [x] I reviewed #152 and #204. This appears related to #152, but the failure occurs after START and creates incorrect authority instead of rejecting candidate-view creation. #204 concerns fresh-process pre-commit recovery, not initial review scope derivation.
- [x] I understand implementation should not begin until a maintainer approves this issue.

## Bug description

A one-file Markdown-only change in a repository with clean, unchanged submodules is classified as a medium-risk executable change. Ordinary review freezes all five submodule gitlinks as deleted/changed, creates authority for the wrong target, and then blocks commit and push.

This is disproportionate for a documentation-only edit and leaves the user in a fail-closed dead end that requires abandoning review authority even though Git itself reports only the intended Markdown file as modified.

## Steps to reproduce

1. Use a repository containing initialized Git submodules whose gitlink pointers match `HEAD` and the index.
2. Modify exactly one tracked `.md` file. In this reproduction, the diff was 2 insertions and 1 deletion.
3. Confirm Git state:

   ```text
   git status --short
    M docs/users/user-guide/compras-y-cuentas-a-pagar/orden-de-compra/ordenes-de-compra-generacion-masiva.md

   git diff --cached --name-status
   # empty

   git ls-files --others --exclude-standard
   # empty
   ```

4. Run `gentle_review inspect`. It correctly reports a one-path workspace target and `action: start`.
5. Run ordinary START.
6. Observe that START creates a medium-risk lineage, selects `review-reliability`, reports six changed files, and attributes executable risk to a clean submodule.
7. Finalize the selected lens.
8. Observe that FINALIZE reports the real one-file target as unrelated with `action: start`.
9. Follow the declared START action once. It reconciles the existing lineage whose frozen projection contains all five submodule paths plus the Markdown file, and returns `next_action: finalize`.

## Expected behavior

- Unchanged gitlinks must remain part of immutable repository identity without entering the changed-path scope.
- A one-file Markdown-only edit must remain a low-risk documentation change.
- START and FINALIZE must resolve the same one-file target.
- The user must be able to obtain a receipt and commit without abandoning stale authority or running an executable-code review lens.

## Actual behavior

Initial START returned:

```json
{
[...]

PR #217, opened 2026-07-22
fix(review): preserve absent gitlinks in snapshots

jacargentina:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved review snapshot accuracy for repositories containing gitlinks or submodules.
  * Preserved multiple absent-but-clean gitlinks in complete snapshots.
  * Correctly retained staged deletions and detected checked-out submodule revisions.
  * Ensured snapshot capture does not unintentionally alter the Git index or worktree.

### DIFF
--- a/lib/review-snapshot.ts
+++ b/lib/review-snapshot.ts
@@ -3,6 +3,7 @@ import { createHash } from "node:crypto";
 import {
 	chmodSync,
 	existsSync,
+	lstatSync,
 	mkdtempSync,
 	mkdirSync,
 	readFileSync,
@@ -153,11 +154,7 @@ export interface CorrectionSnapshotV1 {
 
 const OBJECT_ID = /^[0-9a-f]{40,64}$/;
 
-function runGit(
-	cwd: string,
-	args: readonly string[],
-	environment: GitEnvironment = {},
-): string {
+function gitEnvironment(environment: GitEnvironment): NodeJS.ProcessEnv {
 	const env = reviewGitEnvironment();
 	if (environment.indexFile) env.GIT_INDEX_FILE = environment.indexFile;
 	if (environment.objectDirectory) {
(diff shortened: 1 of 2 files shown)


barbatdev:
Technical review of the diff against `lib/review-snapshot.ts` and `tests/review-snapshot.test.ts` on `main` `19b0ed7`.

The fix is sound. Collecting absent-but-clean gitlinks from the live index before `git add -A` and re-applying them via `update-index --index-info` after is the right call, and the index-immutability assertions prove the live index stays untouched. Two things worth considering before merge:

- **Mixed gitlinks coverage.** Each new test covers one gitlink scenario in isolation. A test with an absent submodule and a present one checked out at a different revision would verify the `lstatSync` branch logic does not interfere between entries. The code handles it correctly, but it is unverified.
- **Large-repo index scan.** `absentLiveIndexGitlinks` reads the full live index via `ls-files --stage -z` and calls `lstatSync` for every gitlink on every snapshot capture. This is O(index size) and runs on the #256 large-repository track. Acceptable given the bug being fixed, but worth noting for the parity performance budget.

Governance gap for the maintainer: the PR body has no `Closes #215`, issue #215 has no `status:approved`, and the PR carries no `type:*` label. The validation checks do not appear to have run.

barbatdev (changes requested):
The gitlink-preservation approach is sound, and the focused RED-to-GREEN coverage proves the core snapshot defect.

Before merge, please add:

- A docs-only regression with multiple absent clean gitlinks, asserting that only the Markdown path enters scope, the route remains trivial, and no review lenses are selected.
- Lifecycle parity coverage proving START, FINALIZE, and pre-commit retain the same candidate tree and path digest.

Renaming the shadowed `gitEnvironment` local is worthwhile cleanup, but it is not blocking. Once the two regression paths pass against current `main`, this should be ready for another review.

barbatdev:
Maintainer revalidation completed against stable gentle-pi v2.2.0 (`49833c8`) and the pinned gentle-ai v2.4.0 runtime.

The underlying #215 bug remains valid, and this patch's snapshot implementation is technically sound: it applies cleanly to stable, the focused snapshot suite passes 13/13, and the index-preservation approach correctly handles the original Pi-side snapshot layer.

However, the production architecture has moved since this PR was opened:

- Ordinary compact START now uses `CandidateViewRegistry` and the provider-owned native snapshot path.
- `captureReviewSnapshot` is only reached by graph-v1 Judgment Day START.
- `captureLiveReviewCandidateBinding` has no production callers.
- The pinned provider still builds its workspace snapshot by copying the index and running `git add -u` without restoring absent gitlinks.

Therefore the previously requested START/FINALIZE/pre-commit parity cannot be satisfied by this diff: the changed functions are no longer on that lifecycle path.

I am closing this as superseded/wrong-layer, not because the implementation quality is poor. Issue #215 should remain open for the provider-owned correction and a compatible gentle-pi pin. The regression fixtures and restoration approach here remain useful input for that work.

PR #266, opened 2026-08-01
feat(release-artifact): decode and bound-extract the provider's signed assets archive

Alan-TheGentleman:
First slice of the consumer chain. Adds the bootstrap decoder that targets the shape gentle-ai declares, without re-authoring any part of that format.

## Prerequisites, verified rather than assumed

The tracker was rebased onto `main` and all three landed prerequisites confirmed present: the single authoritative pin (#262), additive-tolerant gates and projections with mandatory features still exact (#263), and `openspec/specs/package-runtime/spec.md` from the archive (#265). None of them were re-implemented here.

## Two ordering decisions that are the whole point

**Exact path-set equality runs before digesting.** A tree digest computed over the files you found produces the same digest whether or not an extra file exists, because the digest walk only ever revisits declared entries. It structurally cannot notice an addition. So the member set is compared against the manifest first, and only then digested.

That is proven three ways, not asserted: a pure unit test, an orchestration test asserting extraction is never reached once the check rejects, and an end-to-end test that builds two real `.tar.gz` archives with the system tar, one carrying an undeclared file, and watches the production listing path reject it.

**Extraction is bounded before any byte is written.** The manifest's `entries[].size` is enforced during a listing pass ahead of extraction. A digest check after the fact is too late against a decompression bomb.

The unsupported-major failure names the major and happens before any layout inference. Failing closed after guessing a layout is not failing closed.

## Reproducing the provider's digest

The tree-digest preimage is the tag plus a NUL byte, then per entry `path\0type\0mode\0size\0digest\n`. Verified byte-identical against the provider's Go known-vector output rather than reimplemented from the prose description.

## One honest limitation

The real system-tar listing parser is tuned to GNU tar's verbose column format. On a BSD or bsdtar dialect it **fails closed** — it rejects rather than silently misparsing. That is the safe direction, but it is a real gap to harden when the macOS and Windows install paths are exercised in P2b, and it is flagged rather than left to be discovered.

## Tests

42/42 green. `pnpm run check:transaction-runner` clean, with the new module registered in the generator's sources and regenerated rather than hand-edited.

## Rollback

Delete the new lib, runtime, test and fixture files, and revert the one-line sources entry. Nothing else references them.


Alan-TheGentleman:

PR #267, opened 2026-08-01
feat(release-artifact): verify at pin time, mirror offline, and lock what was verified

Alan-TheGentleman:
Second slice of the consumer chain, stacked on #266. Owns trust-order steps 1 through 5, the network boundary; #266 owns steps 6 through 11.

## Where the signature check actually belongs

`scripts/gentle-ai-installer.mjs` documents that the pinned sha256 literals were derived by a human from the signed `checksums.txt`, and install compares against that pin. So minisign verification and repo/tag binding belong to the **pin-time sync job**; install time inherits that trust through the lock.

Shipping a verifier and a trusted key to every end user would be a larger new trust surface for no gain, and it was rejected in design rather than skipped by accident.

## The lock is what makes every PR check offline

It binds release identity, contract major, archive digest, tree digest, canonical entries and generated-output digests. Only an explicit pin-bump job downloads anything.

`scripts/verify-package-files.mjs` loses its hand-maintained ~60-entry `contractHashes` map and reconciles against the lock instead. Three failure modes each fail naming the exact file: a mirror on disk the lock does not list, a lock entry with no file, and a drifted digest.

The offline claim is proven, not asserted: the trusted key is a pending sentinel that throws before any network call, and `verify-package-files.mjs --check` passes reading only the lock and local files.

## A defect this slice found and fixed in the previous one

Running the script end to end surfaced that `lib/release-artifact.ts` and its generated runtime were missing from `requiredPaths` in the package check. Fixed here rather than left for someone to trip over.

## And a defect in this slice, caught in review

The evidence class was carried only on the in-memory sync result. The lock that actually gets checked in, reviewed and later trusted carried **no indication of where it came from**.

That is precisely the failure mode the spec exists to prevent. A bootstrap-derived lock has a placeholder `release.commit` and an `archive.digestSource` reading `signed-checksums.txt` — any reader would take it for a real signed release, and nothing in the file said otherwise.

The lock now persists `evidence.class` and `evidence.signatureStatus`, so a reader holding only the file can still refuse it as pin or final-acceptance evidence. Two tests cover it, including one asserting the checked-in lock itself is labelled bootstrap. That fix is its own commit.

## Status of the checked-in lock

It is `development/bootstrap` evidence and says so. No gentle-ai release exists under this contract yet, so it must be regenerated by a real network `--write` run once the provider release lands and the real minisign key is provisioned.

## Tests

1067 pass, 3 fail. Those three are the known receipt-driven-development-disabled failures in `tests/native-review-parity-runtime.test.ts`, verified identical on clean `main` before this chain started. No regressions.

## One interpretation flagged

The minisign trusted-comment repo and tag convention was defined here because design did not specify an exact wire format for this boundary. It is tested, but it is an interpretation and should be confirmed against the provider's actual signing output before the first real sync.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
         env:
           GENTLE_PI_REQUIRE_NATIVE_BINARY: "1"
 
+      # Offline mirror + capabilities/gentle-ai-release.lock.json reconciliation
+      # (contracts/, docs/gentle-ai/, capabilities/*.semantic.json) runs here,
+      # fully offline. Only scripts/sync-gentle-ai-release.mjs --write, invoked
+      # exclusively by the not-yet-landed pin-bump job, ever downloads or
+      # verifies a minisign signature; no job in this workflow does.
       - name: Verify package contents
         run: node scripts/verify-package-files.mjs
 
--- a/capabilities/gentle-ai-release.lock.json
+++ b/capabilities/gentle-ai-release.lock.json
@@ -1,0 +1,756 @@
+{
+  "release": {
+    "repository": "Gentleman-Programming/gentle-ai",
+    "tag": "v2.2.3",
+    "version": "2.2.3",
+    "commit": "0000000000000000000000000000000000000000"
+  },
+  "contract": {
+    "id": "gentle-ai.release-artifact",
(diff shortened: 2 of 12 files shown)


Alan-TheGentleman:

PR #268, opened 2026-08-01
feat(installer): publish binary and assets as one atomic bundle

Alan-TheGentleman:
Third slice of the consumer chain, stacked on #267. POSIX install path; Windows split provenance is the next slice.

## The half-published bundle becomes unobservable, not just unlikely

`recoverInterruptedPublication` already takes the bundle-validity predicate as a parameter and is already called with `existingSignedBundleMatches`. Extending that **one** predicate to cover the assets tree makes a binary-without-assets bundle invalid *by construction*. No second publish operation, no new crash window.

Two tests prove it rather than describe it: an assets-less existing bundle is rejected for reuse and repaired by a fresh full install, and a crash simulated through a fake `rename` seam refuses to silently restore a backup missing its assets, failing closed and asking for intervention instead.

## Path-set equality before digesting, again

Same reason as #266: a digest walk only ever revisits declared entries, so it cannot notice an added file. `resolveGentleAiAssets` asserts exact path-set equality against the manifest first, then per-file `lstat`, confinement and mode checks, then a whole-set `sameFile` TOCTOU recheck.

## The integrity manifest is extended, never loosened

Five new keys join the existing exact key-count and string-equality discipline in `isCanonicalManifest`. Every asset file gets its own regular-non-symlink check, not just the binary. Symlink rejection, path confinement and TOCTOU detection are untouched.

Threat cases covered: an extra file in the installed tree, a symlinked asset, an asset with mode `0755`, an asset named `install.sh` landing non-executable, TOCTOU replacement mid-verify across the whole file set plus the manifest, and a forged assets digest still rejected by `resolveGentleAiBinary`.

## Where the pinned assets digest comes from

The lock, not a second hardcoded table. Reintroducing a hand-maintained table here would recreate exactly the drift that #262 removed for the binary version.

## Tests

63/63 focused. Full suite 1083 pass, 3 fail — the known receipt-driven-development-disabled failures, verified identical on clean `main` before this chain started. `pnpm run check:transaction-runner` clean, runtime regenerated rather than hand-edited.

`pnpm run test:harness` fails in this sandbox, confirmed by stashing to reproduce identically on the unmodified parent branch. Pre-existing environment condition, same root cause.

## Rollback

Revert the three source files and two test files to the parent tip and delete the new fixture helper. No file from the earlier or later slices is touched.


Alan-TheGentleman:

PR #269, opened 2026-08-01
feat(installer): give Windows the same signed assets with its own binary provenance

Alan-TheGentleman:
Fourth slice of the consumer chain, stacked on #268.

## Windows is not a coverage gap

The release builds linux and darwin binaries only, so Windows builds its binary from Go SumDB source at the pinned tag. But the assets archive has no platform axis, so **Windows downloads and verifies the same signed archive as everyone else**, against the same locked digests.

One atomic bundle records both provenances with their distinct fields: SumDB source-build evidence for the binary, signed-archive evidence for the assets.

## The cross-check observes, it never creates authority

A live `review capabilities` call on the source-built binary cross-checks semantic compatibility against the signed snapshot. It cannot become the source of truth.

That is proven two ways rather than described: the function body contains no write, rename or mkdir call at all, and it is placed strictly before staging, the integrity manifest write and the bundle publish. So any cross-check failure guarantees nothing was staged or published — asserted directly by tests confirming the binary, `integrity.json` and `assets/` are all absent after every failure mode: non-zero exit, oversized output, non-JSON output, and semantic mismatch.

Subprocess discipline: fixed argv, no shell, bounded output, sealed environment, reusing the existing invocation seam.

## Pruning, and a deviation worth naming

`pruneSupersededBundles` runs only after the new bundle's rename succeeds, never touches the live bundle, and its own failure is non-fatal and logged.

The policy implemented is **keep the immediately previous bundle for rollback, prune the rest**. That refines a literal reading of the design's prose, which could be read as removing every non-live version. It follows the maintainer's stated decision and is documented as a deviation rather than applied quietly.

## Offline symmetry

`GENTLE_PI_SKIP_GENTLE_AI_INSTALL=1` skips binary and assets with the same loud disposition and leaves no partially configured bundle directory.

## One task was already green

P2a had already threaded the assets keys through `windowsSourceManifest`, so task 4.6 needed no work. Noted rather than re-claimed as new.

## Tests

64/64 on the focused installer, binary and entrypoint files. One failure in `tests/native-review-cli.test.ts` is environmental: this sandbox has no network, so the pinned binary was never installed. Verified independently by checking out the parent tip in the same worktree and getting an identical 47 pass / 1 fail. The diff did not cause it.

## Rollback

Revert the installer and its test file to the parent tip and delete the new entrypoint test. No library, generated runtime, sync script, lock or CI file is touched.

### DIFF
--- a/openspec/changes/consume-gentle-ai-release-artifacts/apply-progress.md
+++ b/openspec/changes/consume-gentle-ai-release-artifacts/apply-progress.md
@@ -326,3 +326,163 @@ process-level race timing, exactly the same testing pattern already used for the
 
 10/10 tasks in PR 3 (P2a) complete. 28/28 tasks total across P1a+P1b+P2a. Ready for `sdd-verify`, or for
 `sdd-apply` to continue with PR 4 (P2b) in a fresh batch, based on this branch (`feat/release-assets-install`).
+
+---
+
+# PR 4 / P2b: Windows split provenance + cross-check + bundle lifecycle
+
+Worktree: `gentle-pi-worktrees/p2b`. Branch: `feat/release-assets-windows`, based on P2a's
+`feat/release-assets-install`. Tasks 4.1-4.9, all complete.
+
+## Scope discovery: D5's read-side and manifest threading were already complete
+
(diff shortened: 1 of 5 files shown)


Alan-TheGentleman:

PR #270, opened 2026-08-01
feat(baselines): generate the required floor without letting it shrink silently

Alan-TheGentleman:
Fifth slice of the consumer chain, stacked on #269.

## The paradox, and how it resolves

The spec wants `REQUIRED_*` **generated**. The plan wants the floor **frozen**, so a removal stays visible. Those read as contradictory.

**Monotone regeneration** resolves it. `--write` may ADD a required name. A name disappearing from the snapshot's required block makes generation **fail naming it**, and demands a hand-authored deletion from the generated file. A floor that silently follows the provider downward is not a floor.

The mechanism is a self-referential fixed point: the generator regex-parses its own previously checked-in output as the baseline and unions it with the current snapshot. Un-requiring something needs a visible edit to a file marked "do not edit" — and that diff is the review signal.

Proven twice: as a pure-function test, and as a real subprocess run against a shrunk snapshot fixture, which fails with the exact operation name and the instruction to hand-edit.

Same discipline for mandatory features: an advertised one absent from Pi's supported set fails naming it and is never auto-added. Pi's supported set is a decision, not a mirror.

## The 12 frozen rows moved verbatim

They describe already-released immutable versions and must never be recomputed. Their explanatory prose survives as `notes` rather than being dropped.

## Two interpretations flagged rather than quietly closed

**The task text says 13 of 17 capability flags map from the snapshot via `review.<op>`. In practice only 5 map by exact operation name.** The other 8 predate the negotiated contract and legitimately carry forward from history. That gap is documented rather than papered over with a lookup that would look derived but wouldn't be.

**`REQUIRED_SCHEMAS` and the mandatory-feature constants stay hand-authored.** The semantic snapshot mirror carries no `schemas` array, and mandatory support is a client-code decision rather than something to mirror.

Both deserve a reviewer's judgement, which is why they are here rather than in a commit footnote.

## Tests

19 new, all green. The 7 pre-existing capability-contract regression locks still pass. `--check` clean, `check:transaction-runner` clean with six generated modules now in sync.

Full suite 1102 pass, 1 fail — the environmental no-network failure documented since #267, confirmed identical against the unmodified parent tip.

## Rollback

Delete the five new files, revert the two library imports (the original hand-authored constants are fully preserved in the diff's removed lines), the sources entry, the CI step and the package scripts. No decoder, sync, installer or assets trust code is touched.


Alan-TheGentleman:

PR #271, opened 2026-08-01
feat(skills): vendor shared skills read-only with explicit Pi overlays

Alan-TheGentleman:
Sixth slice of the consumer chain, stacked on #270.

## The comparison is the gate, not the allowlist

Nine candidates were compared file by file. **Seven admitted, two rejected.**

Admitted first tier, all five: `comment-writer` (one persona-safety sentence as the Pi delta), `work-unit-commits` (Pi was **stale** — gentle-ai already had content Pi lacked, so this catches Pi up), `branch-pr` and `chained-pr` (only the pre-existing collision-prefix `name:` field differs), `cognitive-doc-design` (byte-identical).

Admitted second tier, two of four: `skill-improver` (seven bounded, nameable deltas) and `skill-registry` (byte-identical).

**Rejected: `judgment-day` and `skill-creator`.** Both diverged into near-total rewrites. Expressing that as an overlay would need a single anchor spanning almost the entire body, which is functionally the full-fork shape the overlay design exists to avoid. No repository name leaks in either — they are simply too far apart to pretend otherwise.

`issue-creation` is excluded entirely: no vendoring, no gate, no comparison. The two copies describe **different repositories** by design. That is not drift to reconcile.

## The failure this actually prevents

A hand-edited vendored file is caught by the drift gate. But the one that matters under a fast provider cadence is the **stale anchor**: upstream rewords a section, the Pi delta's anchor no longer matches, and the delta silently disappears from the skill agents load.

That fails loudly with `edit the overlay, not the vendored file: skills/_vendor/<name>/overlay.md`. A missing anchor and an ambiguous one fail identically, because the remediation is the same.

## One deviation you should know about

**The release archive contains no skill files.** That is by design — the first published archive carries contracts, snapshot and docs only, per the decision to keep the frozen surface small.

So the vendored bodies here were seeded as manually cross-checked byte-for-byte copies from the gentle-ai checkout, not from a live sync run. Re-syncing an admitted skill needs a human re-copy until a git-based skill sync is wired. Flagged as a follow-up rather than presented as automated.

## Tests

15 new, all green. `--check` clean across all seven skills. Full suite 1117 pass, 1 fail — the environmental no-network failure documented since #267.

## Rollback

Each admission is its own commit, and the generator iterates the manifest rather than a hardcoded list, so reverting one removes only that skill. Reverting the whole PR restores every `SKILL.md` and deletes the generator, vendor tree and tests. Nothing from the earlier slices is touched.


Alan-TheGentleman:

PR #272, opened 2026-08-01
feat(coverage): warn on unbound provider phases, fail on unknown Pi bindings

Alan-TheGentleman:
Seventh slice of the consumer chain, stacked on #271.

## The gate is asymmetric on purpose

**A provider-declared phase with no Pi binding WARNS and does not fail CI.** Under the provider's release cadence new phases appear regularly, and blocking the build for a phase nobody has decided to implement yet trains people to ignore the gate. A gate that cries wolf gets muted, and a muted gate protects nothing.

**A Pi binding naming no declared phase, and not listed as Pi-only, FAILS.** That is not upstream moving — it is an inconsistency inside Pi's own configuration, and it should break the build.

This supersedes task 7.1's wording, which said the forward direction fails. The supersession is annotated in `tasks.md` and explained in `apply-progress.md` so it reads as a decision rather than a slip.

The two directions are disjoint code paths: `missingBindings` only ever reaches `console.warn`, and `process.exit(1)` is gated strictly on `unknownBindings.length > 0`. Verified end to end against the real tree, which exits 0 with full coverage.

`issue-creation` is never flagged. It is repo-identity content, not drift.

## The evidence harness Wave 1 will consume

`tests/evidence/**` implements the S-versus-R ledger with one invariant: **a bootstrap record can never be relabeled release evidence**, and a release record requires live signed-release inputs. It builds on the `evidence.class` discipline the lock already carries in #267 rather than inventing a parallel notion.

Wave 1 consumes this rather than reimplementing it. That is the point of putting it here.

## One scope decision, forced by the real tree

The reverse check covers `assets/agents/**` only. Reconciling the real tree against the declared six-item Pi-only list showed Pi's chain files include three names with no matching declared phase and no authorization to add them. Extending the check to chains would have required inventing entries. Documented rather than quietly widened.

## CI

The offline checks consolidate into one integration gate: mirrors and lock, baselines, skill overlays, phase coverage. All of it runs with no network.

## Tests

28 across the two new areas, all green. Full suite 1132 pass, 1 fail — the environmental no-network failure. A second environmental symptom in `tests/runtime-harness.mjs` was checked and reproduces on the clean parent tip too, so neither is introduced here.

## Rollback

Reverting the five commits restores the parent state exactly. No file or behavior from the earlier slices is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,28 +35,36 @@ jobs:
         env:
           GENTLE_PI_REQUIRE_NATIVE_BINARY: "1"
 
-      # Offline mirror + capabilities/gentle-ai-release.lock.json reconciliation
-      # (contracts/, docs/gentle-ai/, capabilities/*.semantic.json) runs here,
-      # fully offline. Only scripts/sync-gentle-ai-release.mjs --write, invoked
-      # exclusively by the not-yet-landed pin-bump job, ever downloads or
-      # verifies a minisign signature; no job in this workflow does.
-      - name: Verify package contents
-        run: node scripts/verify-package-files.mjs
-
-      # Generated NATIVE_CLI_CONTRACTS row + REQUIRED_* floors (D7): offline,
-      # reads only the checked-in semantic snapshot, lock, and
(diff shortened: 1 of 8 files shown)


Alan-TheGentleman:

PR #273, opened 2026-08-01
feat(release): prepare pin-bump PRs automatically, and never merge them

Alan-TheGentleman:
Final slice of the consumer chain, stacked on #272. Completes the tracker at 68/68 tasks.

## The bot prepares evidence. The human decides.

**The bump PR is never auto-merged, even on green.** Merging automatically would change the binary users install with nobody looking.

That is enforced structurally, not by intention: `buildPullRequestArguments` returns exactly ten fixed tokens with no `--auto`, `--admin` or merge verb; the script contains exactly one `gh` invocation in its whole source, `pr create`; the workflow has one job and no merge step. The only occurrences of the phrase "auto-merge" anywhere in this diff are comments forbidding it and the PR body telling the reader the same.

## The tag is hostile input

It arrives from an external event, so it is validated against an anchored `^v\d+\.\d+\.\d+$` **before any use**. A space or dash anywhere fails the pattern, so no extra CLI flag can be smuggled through it.

Proven twice: a table-driven suite covering ten shell-metacharacter payloads and five `--head`/`--base` injection attempts, and live — `--tag 'v1.2.3; touch /tmp/pwned-marker'` is refused with the tag echoed back and creates no file. Verified independently of the implementing run.

The base is the hardcoded literal `main`, never implicit and never derived from the payload. The head branch comes only from the validated tag.

## Digests are written only after the signature verifies

The sync script owns minisign verification and repo/tag binding, and runs first. A test simulates verification failure and asserts baselines, overlays, commit, push and `gh` never run afterward. A bot that trusts a downloaded checksum without verifying its signature has automated a supply-chain hole rather than closed one.

## The receiver is inert, and that is correct

GitHub fires `repository_dispatch` only for a workflow already on the **default branch**. This receiver cannot run until it merges to `main`, and future bump PRs target `main`, never a temporary tracker.

That is stated in the workflow's own comments so nobody later "fixes" it.

## What still has to happen before this runs for real

This chain merges to `main`; the provider publishes a real signed release, since the trusted key is still a pending sentinel; the provider dispatches with the tag; and the bot token's permissions get verified against real GitHub Actions, which this offline sandbox cannot do.

Those are activation steps, not code gaps.

## Tests

33 new, all green. Full suite 1165 pass, 1 fail — the environmental no-network failure carried since #267. The real subprocess path was additionally exercised against a throwaway git repo with a stub `gh`, confirming the exact `--base main --head bump/gentle-ai-v9.9.9` tokens reach it.

## Rollback

Delete the script, its tests and the receiver workflow, and revert one CI comment. Nothing from the earlier seven slices is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,13 @@ jobs:
       #      hand-edited vendored file and a stale overlay anchor before
       #      either reaches skills/<name>/SKILL.md.
       #
-      # Only scripts/sync-gentle-ai-release.mjs --write (the not-yet-landed
-      # pin-bump job, P4) ever downloads or verifies a minisign signature; no
-      # step in this workflow does.
+      # scripts/sync-gentle-ai-release.mjs --write only ever runs in the
+      # pin-bump job of .github/workflows/gentle-ai-release-received.yml
+      # (design D1/D6/P4) -- no step in THIS workflow downloads a gentle-ai
+      # release or verifies a minisign signature. (This job's own `pnpm
+      # install` above still reaches the npm registry for dependency
+      # installation; that is a separate, pre-existing surface unrelated to
+      # the gentle-ai release trust boundary this design adds.)
       - name: Verify offline checks (mirrors/lock, phase coverage, baselines, skill overlays)
         run: |
           node scripts/verify-package-files.mjs --check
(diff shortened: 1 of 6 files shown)


Alan-TheGentleman:

PR #274, opened 2026-08-01
feat(review): capture reviewer results against the exact provider-issued slot

Alan-TheGentleman:
First slice of Wave 1, the #2028 host behavior. Independent of the release-artifact chain; only W3 depends on that foundation.

## The defect was narrower than it looked

`providerReviewerProjection` **already validated every binding field** — lineage, authority revision, target, base and candidate trees, changed-path manifest hash, lens, order, subject hash, including argument-token equality and duplicate detection. Then it threw the result away.

So this is an **extraction**, not a new validator. Writing a second one would have created exactly the competing provider authority the plan forbids. `deriveCaptureSlots` is now the single six-field validator; both the projection and the new FINALIZE capture phase call it.

Behavior preservation was verified by running the pre-existing 320-test suite unchanged across the refactor step alone, before anything new was added.

## Pi transports, it does not interpret

Argument tokens stay discrete argv elements, byte-identical, no shell, fixed executable. Capture cwd is `candidateView.root`. Tokens carrying `--repository-context` alongside an explicit cwd are rejected, and empty or non-string tokens are refused.

An admitted-manifest path containing `..`, an absolute path, or an executable-looking name is **forwarded byte-identical with zero Pi-side filesystem access**. Pi is not in the business of deciding what a provider path means. A manifest carrying both `path` and `reference` is refused outright.

## Dead code, and a correction to the task text

The task said to drop `resultFiles` and `lensResults` from `NativeFinalizeRequest`. **That would have broken the legacy plain-CLI client**, which genuinely stages and passes them as `--result` argv and is still exercised by roughly twenty fixture-compatibility tests.

The provably dead staging was in the negotiated production client only, where documents were written to a 0600 tmp file that its argv never referenced. That is what was deleted. The shared type keeps the fields, with a comment naming who still uses them.

## One thing found along the way

`captureResult()` existed on the production client but was **never declared on the `NativeReviewCli` interface**, so no interface-typed caller could reach it. Declared here.

## Transport selection is per run

Every slot carrying `path` selects the artifact-file list in ascending order. Any `reference`, or any already-committed slot, selects the captured-results flag. The retired `--result` is never emitted from the negotiated path.

## Tests

17 new pure-function tests plus new bijection, cwd, verbatim-token, transport-selection and documentation-like-path coverage. 1022 pass, 3 fail — the receipt-driven-development-disabled failures, confirmed identical against a stashed baseline. `check:transaction-runner` green with the runtime regenerated rather than hand-edited.

Receipt-driven development stays disabled throughout; nothing here starts, recovers, retries or reclaims review authority.

## Rollback

The two feature commits revert together, restoring the prior FINALIZE argument mapping. The provider binary is untouched.


Alan-TheGentleman:
This capture-slot extraction is useful, but this PR is not mergeable as the new Pi provider path. Gentle AI provider tasks now require a Go-issued opaque host task, while this branch still participates in Pi-owned reviewer, refuter, or validator document construction.

The canonical parity contract is now #311. Its provider prerequisite is explicit Pi or generic host-relay support from Gentle AI. Once that released contract exists, the slot-validation extraction here can be split into a narrow transport slice, but no part of this PR may preserve local role semantics, use the OpenCode relay, or pass Pi as `--agent opencode`.

Keep this branch blocked until #311 P1 through P4 have a released provider schema and capability to target. Then rebase only the reusable capture-slot work and delete the local semantic FINALIZE path.

Alan-TheGentleman:

PR #275, opened 2026-08-01
feat(review): safe diagnostics, one bounded relaunch, and lost-output recovery

Alan-TheGentleman:
Second slice of Wave 1, stacked on #274.

## Diagnostics are an allowlist, because a wrong default leaks user data

`decodeSafeAdmissionDiagnostic` returns a value only for its exact known shape and `undefined` for everything else. Denylist and sanitize-then-forward were both rejected in design: a denylist's wrong default is precisely how private paths and user content end up pasted into an issue report.

Rejection cases proven: unknown code, extra key, missing required key, malformed finding id, absolute or home-directory location, `..` and backslash escapes, control characters and newlines, over-length reason or location, untrimmed raw prose, and non-object input.

## One relaunch, and the controller does not launch it

One grant per slot. An identical six-field STATUS reoffer plus a safe diagnostic returns `blocked: capture-relaunch-required` — the controller hands back a blocked envelope and the parent decides. An already-granted record is terminal `exhausted`. Any field mismatched, or an undecodable diagnostic, is terminal `unavailable`.

Unreplayability stores `sha256(rejectedDocument)`, never the bytes. Retaining candidate-derived content across turns to compare later would be its own problem. Membership is checked before **every** relaunch capture attempt, not only inside the rejection handler, because re-entry resubmits through the normal per-slot loop.

## Lost output: ambiguity fails closed

Proof that a capture committed is that the slot is **no longer offered** under an unchanged lineage id and revision. When that holds, mark committed, do not recapture, do not rerun the lens.

When the proof is absent, take only the declared reconciliation action. **When STATUS is ambiguous, it is not proven.** Guessing there means either a duplicate capture or a silently lost result, and neither is acceptable.

`reconcileNativeMutationFailure` gained an optional prefetched-status parameter so the "one fresh STATUS query per failure" rule holds instead of double-querying.

## Stale grants are structurally impossible

Cleanup fires on all-slots-settled, on FINALIZE terminal, and on session shutdown. But the guarantee that matters is not the cleanup: `slotKey` embeds the authority revision and subject hash, so a grant from a superseded authority cannot key into a current lookup at all. The test asserts that as a property, not as a scenario.

## Receipt-driven development stays disabled

Verified independently across the whole Wave 1 diff: no start, recover, reclaim, reset or reconcile-authority call was introduced. The only native calls are capture and a read-only status query.

## One thing worth knowing about the pinned provider

Today's pinned version has no real admission-diagnostic wire field, so every capture failure reports an unknown outcome regardless of cause. The rejected branch is therefore distinguished by first checking for a decodable diagnostic-shaped property, before the generic lost-output path. That is a defensive decode against the frozen shape and should be revisited at the next pin.

## One placement deviation

The integration tests went into the routing test file rather than the recovery one the task named, because the recovery file has none of the git-repo, candidate-view or manifest fixtures these need, while the routing file already hosts every W1 capture fixture. Recorded in `tasks.md`.

## Tests

35 focused, all green. Full suite 1046 pass, 3 fail — the receipt-driven-development-disabled failures, byte-identical to a stashed baseline. `check:transaction-runner` green.

## Rollback

The two feature commits revert together. W1's capture-only behavior remains valid standalone.


Alan-TheGentleman:

PR #276, opened 2026-08-01
test(review): block #1819/#1915 acceptance until the evidence foundation ships

Alan-TheGentleman:
Final slice of Wave 1, stacked on #275. Completes #2028 host behavior.

## This PR intentionally adds one failing test. Please do not silence it.

The final-acceptance check for #1819 and #1915 runs against the foundation's evidence harness, which lives in the sibling change `consume-gentle-ai-release-artifacts` and has not shipped into this repository yet.

**Absent qualifying evidence must block.** A silent skip would record a false success for behavior nobody verified — the exact failure this spec exists to prevent. So the test fails, loudly, with a message that explains it is the correct fail-closed state, names the two conditions that turn it green, and asks the reader not to add a skip or delete it.

It fails in **both** directions: if the harness is missing it fails saying so, and if the harness appears but the real check has not been wired, it fails demanding that work. It cannot pass by accident.

Its failure is scoped to itself via a dynamic import rather than a top-level one, so it does not abort the rest of the file.

Confirmed the only intentional failure Wave 1 adds: the suite reports 1062 pass and 4 fail — this one plus the three pre-existing receipt-driven-development-disabled failures, byte-identical to the pre-W3 baseline.

## Pi asserts outcomes. It never rebuilds provider algorithms.

#1819 and #1915 are provider-owned: corrected-delivery topology and receipt discovery, retry-successor authority and evidence-graph validation. A static guard proves Wave 1 introduced no delivery-topology or authority-graph symbols into `lib/` or `extensions/`.

The guard scans **declaration sites only**, not bare substrings, because a naive scan false-positived on two legitimate strings: a `--squash` git-commit flag in an allowlist, and an echoed provider capability name. Matching those would have made the guard noise.

## #2074 and #910 are proven by asserting absence

Prose cannot fail, and a fixture would manufacture the false coverage claim the plan warns about. So the tests assert the surfaces genuinely do not exist: no Claude user-registry write path, no PowerShell fallback ladder, no fixture or journey named for either issue, and the coverage enumeration lists both as no-action rather than as covered.

All passed on first run. The surfaces really are absent.

## The receipt-driven-development boundary is asserted, not assumed

Global and effective mode is checked as `off` before and after the file runs, against the real installed binary rather than a fake, and delivery evidence reports `disabled/unmanaged`.

## Rollback

Delete the two new files. Nothing depends on them, and the foundation harness is untouched because it does not exist here.


Alan-TheGentleman:

PR #281, opened 2026-08-05
feat(skill-registry): index npm-package skills (node_modules/*/skills)

ChaoXu1997:
## Problem

`extensions/skill-registry.ts` scans only **loose** skill directories (`~/.pi/agent/skills`, `<cwd>/.claude/skills`, `~/.trae/skills`, …) and **misses npm-package-installed skills** — the ones living at `~/.pi/agent/npm/node_modules/*/skills/`.

These include gentle-pi's own skills (`gentle-ai-*`, `comment-writer`, `cognitive-doc-design`, `release`, `work-unit-commits`), plus `context-mode`, `pi-lens`, `pi-mcp-adapter`, `pi-subagents`, `pi-intercom` — roughly **26 skills** that pi loads into the agent's `<available_skills>` but the registry never indexes.

## Impact

Subagent delegation consumes `.atl/skill-registry.md` to discover and pass skill paths. Because npm-package skills aren't indexed, a delegated subagent can't see them — e.g. a task delegated to a subagent that needs a gentle-ai / context-mode / pi-lens skill gets `skill_resolution: none`, even though the skill is installed and available to the parent agent.

On my machine the registry went from **53 → 79** skills after this patch (26 package skills correctly indexed under a new `package` scope).

## Fix

1. **`discoverNpmPackageSkillDirs()`** — scans `~/.pi/agent/npm/node_modules/*/skills/` and returns the existing per-package skill dirs.
2. Wire it into **`regenerateRegistry`** (scan list) and **`startSkillRegistryWatcher`** (watch dirs), so adding/removing a package skill refreshes the registry.
3. **`scopeForPath`** gains a `"package"` scope for `node_modules` paths, so the registry distinguishes `project` / `user` / `package` sources.

```diff
+async function discoverNpmPackageSkillDirs(): Promise<string[]> {
+	const home = homedir();
+	const nodeModules = join(home, ".pi/agent/npm/node_modules");
+	if (!(await pathExists(nodeModules))) return [];
+	… // readdir node_modules, collect <pkg>/skills dirs that exist
+}

 const existingDirs = await uniqueExistingDirs([
 	...projectSkillDirs(cwd),
 	...userSkillDirs(),
+	...(await discoverNpmPackageSkillDirs()),
 ]);

 function scopeForPath(cwd, path) {
+	const cleanPath = comparablePath(path);
+	if (cleanPath.includes(`${sep}node_modules${sep}`)) return "package";
 	… // existing project / user logic
 }
```

[...]

PR #285, opened 2026-08-06
feat(skill-registry): index pi.skills project-declared skill dirs

ChaoXu1997:
## Problem

`projectSkillDirs(cwd)` scans a fixed list of convention directories (`cwd/skills`, `cwd/.pi/skills`, `cwd/.claude/skills`, …). It does **not** read `package.json`, so projects that declare their skills via the `pi.skills` multi-path array (e.g. `"./penpot-design/skills"`) are **never indexed** — those skills are absent from `.atl/skill-registry.md` and therefore invisible to delegated subagents.

## Fix

Add `piSkillsDirs(cwd)`:
- reads `<cwd>/package.json` → `pi.skills` (string array)
- resolves each entry relative to `cwd`
- confines results to the project (rejects entries that traverse outside `cwd`)
- wired into both `regenerateRegistry` and `startSkillRegistryWatcher`

`findSkillFiles` already descends one level (`root/*/SKILL.md`), which matches the `pi.skills` layout (each entry points to a directory of skill subdirs), so no recursion change is needed. Paths under `cwd` are already classified `"project"` by `scopeForPath`, so scope is correct without changes.

## Verification

- `node --experimental-strip-types --check` passes.
- On a project with `pi.skills: ["./penpot-design/skills", "./penpot-utilities/skills", "./penpot-build/skills"]`, all 16 declared skills (5+6+5) are discovered after the patch (0 before).

## Scope / relation to #281

Independent of #281 (npm-package scan). Both PRs touch the same two scan-list lines, so merging both produces a trivial 1-line-per-list conflict that resolves to `...discoverNpmPackageSkillDirs(), ...piSkillsDirs(cwd),`.

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering skills from paths declared in a project’s package configuration.
  * Valid skill directories are included when refreshing the skill registry and monitoring for changes.
  * Invalid, empty, malformed, or unsafe paths are ignored.

### DIFF
--- a/extensions/skill-registry.ts
+++ b/extensions/skill-registry.ts
@@ -74,6 +74,32 @@ function userSkillDirs(): string[] {
 	];
 }
 
+async function piSkillsDirs(cwd: string): Promise<string[]> {
+	// Read the pi.skills multi-path array from the project package.json so projects
+	// that declare skills via package.json (e.g. "./penpot-design/skills") are indexed,
+	// not only fixed convention dirs like cwd/skills or cwd/.pi/skills.
+	const pkgPath = join(cwd, "package.json");
+	if (!(await pathExists(pkgPath))) return [];
+	let pkg: { pi?: { skills?: unknown } };
+	try {
+		pkg = JSON.parse(await readFile(pkgPath, "utf8"));
+	} catch {
+		return [];
+	}
+	const skills = pkg?.pi?.skills;
+	if (!Array.isArray(skills)) return [];
+	const out: string[] = [];
+	for (const entry of skills) {
+		if (typeof entry !== "string" || entry === "") continue;
+		const candidate = join(cwd, entry);
+		// Confine to the project so a malformed entry cannot traverse outside cwd.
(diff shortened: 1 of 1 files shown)


barbatdev:
Maintainer revalidation completed against stable/current gentle-pi v2.2.0 (`49833c8`).

The project-declared skill discovery gap is valid and is now covered by approved issue #369. This patch applies cleanly and the existing skill-registry suite passes 17/17, but it adds no regression coverage for the new behavior.

The proposed literal `package.json` parser would duplicate only part of Pi's resource-loader semantics. Pi manifests support glob patterns, exclusions, package filtering, disabled resources, and package-local resolution that should not be independently recreated by the registry.

We have selected PR #281 as the single consolidation vehicle for #369. It has been asked to consume Pi-resolved skill records and absorb the valid project-declared outcome from this PR.

I am closing #285 as superseded to avoid maintaining two overlapping resource authorities. This does not reject the reported gap or the contributor's analysis; that work remains represented in #369 and should be carried into the consolidated implementation.

Issue #256, opened 2026-07-31
feat(parity): adopt Gentle AI post-v2.2.2 behavioral parity


Alan-TheGentleman:
### Pre-flight checks

- [x] I searched open and closed issues for Gentle AI parity, candidate decline, large-repository review failures, SDD attempt authority, provider-defect handoff, issue workflow discovery, privacy review, and skill fingerprints.
- [x] This is not a duplicate of #158 or #172. Those delivered earlier provider generations. Existing #196, #242, and #254 remain focused child issues and are referenced rather than duplicated here.
- [x] I understand implementation requires `status:approved` and that each child work unit must remain independently reviewable.

### Problem

gentle-pi 2.1.2 intentionally pins the stable Gentle AI v2.2.2 runtime. Gentle AI `main` has since added user-visible behavior for candidate-scoped review decline, bounded SDD attempts, reviewer recovery, large-repository admission, provider-defect reporting, repository issue-policy discovery, privacy review, and skill-registry invalidation.

Updating only the binary pin would not provide experience parity. Several behaviors require explicit Pi adapter, controller, orchestrator, skill, and runtime-harness changes. Other behaviors are provider-owned and should be consumed through a stable release with package-local integration proof rather than reimplemented in TypeScript.

### Frozen audit baseline

| Component | Audited revision |
| --- | --- |
| Gentle AI base | `v2.2.2` / `09c2d8b6369860058aa44704df76b1b33c745675` |
| Gentle AI main | `0126077c23e8e542bf79f98dc427fbed96086517` |
| gentle-pi | `v2.1.2` / `3b6b3d2183dbbc4d45b16a1a0f127728c0a2435c` |

The audit covered all 46 Gentle AI commits in that range.

### Upstream release prerequisites

The provider must not be repinned from the audited main revision yet:

- six Windows full-suite shards and one Ubuntu E2E check are failing;
- runtime emits `candidate_declined/unmanaged`, while the referenced operation schema still permits only `disabled/unmanaged`;
- no stable v2.2.3 tag, immutable release assets, signatures, archive/binary digests, or Windows SumDB checksum exists.

Gentle AI must first reconcile the runtime/schema value, restore required CI, and publish a stable release.

#### 1. Large-repository reviewability

Three independent Pi limits must be resolved. Fixing one does not unblock the complete flow:

- #254: negotiated native status/process output can exceed the client output buffer;
- #196: candidate-view construction buffers the complete recursive Git tree;
[...]

PR #291, opened 2026-08-09
test(devbinary): derive the start fixture's projection from the real repository

Alan-TheGentleman:
Part of #256, track 8. This does **not** close it, and it does not make the devbinary axis green. Draft on purpose: what is left red after this is the real parity work.

## What this changes

`unrelatedStartTargetStatus()` hardcoded `"b".repeat(40)` as every tree and an empty path list. `assertNativeStartCandidateBinding` compares exactly `baseTree`, `initialReviewTree`, `currentCandidateTree` and sorted `paths` against the candidate view gentle-pi derives from the real workspace, so those placeholders could never match. Every start-driven journey died on `candidate-target-projection-drift` before the binary under test was asked anything.

The fixture now mirrors `lib/review-candidate-view.ts` exactly: read HEAD into a scratch `GIT_INDEX_FILE`, stage everything, write that tree, and take paths from `deriveChangedPathManifest`. Not a placeholder loose enough to slip past the guard, the same values the guard computes.

## Before and after

Against `v2.4.0-rc.1`, `main`, and one more build, all three: 2 pass / 3 fail, all three failures `candidate-target-projection-drift`.

After: the drift is gone and the real binary runs. The three remaining failures are behavioral, and they are #256's content, not this file's. The clearest one is tier 0: it expects a structured start result carrying `lenses_required: false` and a `hint`, while gentle-ai answers with a non-zero exit and the hint on stderr:

```
"stderr": "Error: the candidate has no pending changes; already-committed work can be reviewed by
rerunning review start with --base-ref <commit> naming the base to compare against"
```

Worth noting alongside that: no released capability row (2.2.0 through 2.2.3) sets `hint: true` or `riskEvidence: true`. The devtest spoofs both on in `test.before`, so those two journeys have always been describing a provider generation that no pinned release actually is.

---

## For whoever picks this up

This branch exists because of a report on Discord: running the devbinary tests against the 2.4.0-rc.1 binary shows placeholder OIDs in the projection (`baseTree "bbbb..."`, `pathsDigest "sha256:aaaa..."`) and the tests fail at binding before START. That report is accurate about what it saw and worth reading carefully, so here is the whole picture in one place.

**The placeholders were never the binary's.** `journeyNative()` overrides `targetStatus` with the fixture this PR fixes, and `bridgeAdapter()` intercepts the `version` argv to return a fake `BRIDGE_VERSION = "9.8.7"`. On that path gentle-ai's status is never called and the version handshake never happens. Both are deliberate: the header comment scopes this file to the plain-CLI decode path, not the negotiated status contract.

**It was never RC-specific.** Identical 2 pass / 3 fail against `v2.4.0-rc.1`, against gentle-ai `main`, and against a third build. Three different binaries, one result, because the guard rejected the fixture before any of them mattered.

**Real installs are not affected.** `lib/native-review-cli.ts` refuses a provider whose `packageVersion` differs from the pin and names both versions in the error. Only the `GENTLE_AI_DEV_BINARY` harness gets past that, precisely because it spoofs the version reply.

**Running these tests against an RC is a declared non-goal.** #256 lists "Pinning an RC or mutable provider main" under Non-goals. The post-2.2.2 uplift is that issue, and it is gated on a stable upstream release. So a finding like this belongs as a comment on #256 under track 8 rather than as a new issue, which is exactly the call the reporter made when they asked first.

**What is actually left.** After this PR the harness stops lying about the projection, which is the precondition for using it to measure parity at all. The remaining three failures are the gap between what gentle-pi expects of a post-2.2.2 provider and what gentle-ai currently does. Closing them is track 8 work: repin the stable release, vendor its published schemas and fixtures, and consume the contracts gentle-ai ships instead of maintaining a parallel transport layer here.

### DIFF
--- a/tests/devbinary/native-review-parity.devtest.ts
+++ b/tests/devbinary/native-review-parity.devtest.ts
@@ -6,6 +6,7 @@ import { join } from "node:path";
 import test from "node:test";
 import type { ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent";
 import { createGentleAiExtension } from "../../extensions/gentle-ai.ts";
+import { deriveChangedPathManifest } from "../../lib/review-candidate-view.ts";
 import {
 	NativeReviewCliV214,
 	createNodeExecFileAdapter,
@@ -77,7 +78,7 @@ function journeyNative(binary: string): NativeReviewCli {
 		reviewMode: (request) => bridge.reviewMode(request),
 		async targetStatus(request) {
 			return request.lineageId === undefined
-				? unrelatedStartTargetStatus()
+				? unrelatedStartTargetStatus(request.cwd)
 				: currentTargetStatusFixture(request.lineageId, request.cwd);
 		},
 	} as unknown as NativeReviewCli;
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman:
Closing this draft as superseded: devbinary parity moved on with the pinned v2.2.x line and the atomic relay work, and the fixture this branch patches no longer reflects the current projection. A fresh slice against current main would be smaller than rebasing this.

PR #342, opened 2026-08-16
fix(review): negotiate status for the pi reviewer transport so the host relay runs

Alan-TheGentleman:
Third field failure on the recovered-lineage defect, after #340 and #341 shipped and the package was regenerated: STATUS recognises the lineage, the Pi relay never runs, no lens is launched, zero mutations.

## Root cause (measured, not inferred): the adapter never named its agent

The provider only emits the materialize-marked host-relay slot when the caller identifies its immutable reviewer runtime. `NativeReviewCliV216.targetStatus` never sent `--agent pi`.

Measured on a faithful reproduction against the live binary — linked git worktree (not the primary checkout), uncommitted tracked modifications, successor from an external `review recover --disposition scope_changed`:

```
agent-less STATUS   -> collect input arguments:
  lineage, expected-revision, target, repository-context, lens, order, subject-hash
  artifact_subject: present     submission: ABSENT

--agent pi STATUS   -> the same input PLUS:
  agent=pi, materialize=true    submission: operation_token=capture-result
```

Through the adapter, before this PR:

```
adapter STATUS decoded -> reasonCode: reviewer_results_required
  arg names: lineage,expected-revision,target,repository-context,lens,order,subject-hash
  reviewHostRelaySlots(inputs).length = 0     <-- the relay gets nothing
```

So `reviewHostRelaySlots()` returned zero for every real lineage, the relay branch in FINALIZE never executed, and no lens was ever launched. Every earlier fix exercised dispatch-shaped slots, which is why all of them passed while the field kept failing.

**The other two hypotheses were measured and ruled out:**
- *Handshake env*: not the cause. `gentleAiProcessEnvironment()` already stamps `GENTLE_PI_REVIEW_RELAY_CONTRACT` on **every** adapter invocation, and the relay stamps it again for its own gentle-ai calls. The refusal the maintainer saw bare-handed does not occur through the adapter.
- *Relay needing a candidate view*: false. `lib/review-host-relay.ts` has zero candidate-view coupling — it consumes provider tokens only. Its materialize leg was also verified to work from any cwd (opaque `repository-context` binds the repository): identical 9890-byte prompt from the candidate worktree and from an unrelated directory.

## Fix

- `NativeTargetStatusRequest` gains `agent?: "pi"`, and `targetStatus` renders `--agent pi` when asked.
- The FINALIZE lane negotiates status for the host transport. The agent is **probed, never assumed**: the pinned 2.2.3 provider refuses the flag (verified: `failure/v2`, `operation_failed`), so a typed transport refusal is remembered per provider instance and the lane falls back to the agent-less status. Nothing regresses for providers without the pi transport.
- The relay lane also hydrates the dispatch binding (best-effort) so a hand-dispatched reviewer on the same lineage still resolves; the relay itself does not need it.

## Deliverable 4 — the typed refusal is surfaced, not swallowed

When a provider refuses the pi transport, the envelope carries `relay_transport: { supported: false, code, message }` and the blocked reason names the code and the provider's own words ("supported immutable review runtimes: ..."). It is never degraded into a generic candidate-view message; a regression test asserts the candidate-view string is absent from that envelope.

## RED evidence

`tests/review-relay-transport-agent.test.ts` — fake provider mirrors the measured live behaviour (materialize slot only when the pi agent is named). All three RED before the fix:
- "the controller must negotiate STATUS for the pi reviewer transport"
- "the transport is probed once" (typed refusal + fallback + surfaced cause)
- probe-once memoization

Battery RED (fix files stash-reverted): `relay materialize slot on an externally recovered lineage` FAILED with "the provider materialize slot never reached the host relay (relayed 0); outcome=reviewer-results-required".

## Battery

New check `relay materialize slot on an externally recovered lineage`:

```
relay materialize slot on an externally recovered lineage  PASS  provider offered the materialize slot to the adapter and it reached the relay verbatim (agent=pi, materialize=true, submission=capture-result); the real materialize leg returned 9092 prompt bytes. Residual gap: the locked-down pi subprocess and the provider submit leg are not executed here (model spend)
```

It drives a linked worktree + external recover through the controller, asserts the provider-issued transport tokens and submission reach the relay verbatim, and runs the **real** materialize leg against the real binary. Residual gap declared in the note rather than faked: the pi subprocess and the submit leg cost model spend and are not executed.

Two existing checks were updated, not weakened: with the transport enabled the correct route for a relay-capable provider is the host relay rather than the blocked capture-result step, so they now accept either provider-correct route and still fail on any leak into evidence ordering.

## Verification (all foreground)

- New unit file 3/3; adjacent suites (host-relay routing, controller routing, hydration gap, recovered routing, parity) 262/262.
- Full suite: 1258 tests, 1243 pass; the 14 failures are the known pre-existing dev-binary-override environment set, name-for-name identical to prior runs.
- `pnpm test:cross-lane`: 16 checks, 0 failed. `check:transaction-runner`: no drift. `check:provider-contract`: pass. `test:harness`: exit 0.

The maintainer's lineage and worktree were never touched; everything above was measured on an independent scratch reproduction.

## Summary by CodeRabbit

* **New Features**
  * Added support for negotiating review-result transport with providers.
  * Review workflows can now fall back automatically when a provider declines the preferred transport.
  * Transport refusal details are surfaced when reviewer results are required.
  * Host-relay execution now supports recovered review workflows and materialized prompts.

* **Bug Fixes**
  * Improved finalization and workspace rebind handling for provider-specific transport differences.
  * Prevented relay attempts when the provider does not support the requested transport.

### DIFF
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -6086,6 +6086,38 @@ async function executeReviewControllerOperation(
 					? reviewHostRelaySlots(negotiatedStatus.nextTransition.collect?.inputs ?? [])
 					: [];
 				if (hostRelaySlots.length > 0 && input.final_evidence === undefined && input.correction_line_forecast === undefined) {
+					// One cost/side-effect forecast BEFORE launch, once per
+					// FINALIZE and never per lens: each host-relay slot runs a
+					// real locked-down `pi` reviewer subprocess against the
+					// user's own model, so an unacknowledged finalize would spend
+					// tokens as a silent side effect of what reads like
+					// bookkeeping. Same acknowledgement shape the adapter already
+					// uses for consequential inputs (committedOnly): the caller
+					// states the cost it accepts, which keeps headless callers
+					// working without an interactive prompt.
+					if (input.reviewer_run_acknowledged !== true) {
+						const forecastLenses = hostRelaySlots.map((slot, index) => slot.lens ?? `slot-${index}`);
+						return {
+							operation: parameters.operation,
+							status: "blocked",
+							outcome: "reviewer-model-run-forecast",
+							reason: `Finalize is about to run ${hostRelaySlots.length} real reviewer model run${hostRelaySlots.length === 1 ? "" : "s"} through the pi host relay (${forecastLenses.join(", ")}), one locked-down pi subprocess per outstanding lens, in the foreground. This spends model tokens on your configured model and provider.`,
+							cost_forecast: {
+								transport: "pi_host_relay",
+								model_runs: hostRelaySlots.length,
+								lenses: forecastLenses,
+								side_effects: [
+									"one locked-down pi subprocess per lens, in an empty scratch directory with every discovery surface disabled",
+									"each captured reviewer result is admitted natively into this lineage's authority",
+									"no candidate file, index, or commit is modified",
+								],
+								model_selection: "user-owned: the relay never sets --model, --provider, or --profile",
+							},
+							mutation_performed: false,
+							mutation_outcome: "none",
+							next_action: "Re-run finalize with {\"reviewer_run_acknowledged\": true} to authorize exactly this reviewer work.",
+						};
+					}
 					// The relay itself needs no candidate view (it consumes only
 					// provider-issued tokens), but a session that dispatches a
 					// reviewer by hand on this same lineage does. Hydrate here too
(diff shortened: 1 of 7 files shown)


Alan-TheGentleman:
## Follow-up: the relay spend is now forecast before it happens (e821a6de)

You were right to stop on this. Measured, not inferred.

**1. What the user saw before the spawn: nothing.** Instrumented the controller around the spawn point on the live reproduction:

```
user-visible output BEFORE the pi spawn: []
ui.confirm asked before spawn: false
relay reached with 9 provider tokens
```

Supporting facts: `REVIEW_HOST_RELAY_PI_ARGV` is `pi --print --mode text --no-session --no-tools ...` with the prompt on stdin — a real model run on the user's own model; the tool's `execute` signature ignores its progress channel (`_onUpdate`); and there is no cost/spend/model-run notice anywhere in `extensions/gentle-ai.ts` or `lib/`. So a plain `finalize` spent tokens silently. Confirmed defect, introduced by enabling the transport in this PR.

**2. Smallest mechanism, following the existing convention.** The codebase already gates consequential inputs with an explicit acknowledgement (`committedOnly`, `acknowledgeUntrustedBundleSource`) and reports refusals as typed blocked envelopes with a `next_action` that names the exact resubmission. I followed that rather than inventing a prompt, and deliberately did **not** use the `ctx.ui.confirm` pattern from `authorizeDestructiveReviewOperation`: it fails closed headlessly, which would break non-interactive finalize entirely.

FINALIZE with host-relay slots and no acknowledgement now returns, spending nothing:

```
status: blocked
outcome: reviewer-model-run-forecast
reason: Finalize is about to run 1 real reviewer model run through the pi host relay
        (review-reliability), one locked-down pi subprocess per outstanding lens, in
        the foreground. This spends model tokens on your configured model and provider.
cost_forecast:
  transport: pi_host_relay
  model_runs: 1
  lenses: [review-reliability]
  side_effects:
    - one locked-down pi subprocess per lens, in an empty scratch directory with every discovery surface disabled
    - each captured reviewer result is admitted natively into this lineage's authority
    - no candidate file, index, or commit is modified
  model_selection: user-owned: the relay never sets --model, --provider, or --profile
mutation_performed: false   mutation_outcome: none
next_action: Re-run finalize with {"reviewer_run_acknowledged": true} to authorize exactly this reviewer work.
```

One forecast per finalize covering every outstanding lens — never one per lens. The tool description documents the field.

**RED evidence:** `finalize forecasts the reviewer model run once and spends nothing until it is acknowledged` failed first with *"no reviewer model run may start before the forecast is acknowledged"*; the other three transport tests also went red because the parser rejected the new key until it was added.

**Scope note (not changed, reported):** `executeProviderRoleVectorCollection` spawns Go-owned pi processes for refuter/validator vectors and is equally unannounced. That path predates this PR and I did not touch it — flagging it rather than expanding scope silently. Say the word and I will gate it the same way.

**3. Battery notes now state the guard explicitly**, so a future reader cannot read the two-route acceptance as permission:

- `recovered routing: finalize offers capture-result, never evidence ordering` — "finalize followed the provider transition into the pi host relay for the outstanding reviewer result. Accepting either provider-correct route (relay when the provider admits the pi transport, blocked capture-result otherwise) is NOT a relaxation of the evidence-ordering guard: this check still fails on any evidence-first-ordering leak in the envelope, and on any route that is neither of the two"
- `recovered field flow: linked dirty worktree, finalize-first dispatch` — "...resolved, on whichever provider-correct route was offered, and the envelope still carries no evidence-first-ordering leak; residual gap: the battery cannot age a lineage by days or span OS processes, only a fresh registry"

Both assert the `evidence-first-ordering` string is absent from the envelope; the field-flow check gained that assertion in this commit.

**Verification (all foreground):** transport suite 4/4; relay-routing suite 7/7 (updated to authorize the run, since their subject is routing — the gate has its own test); full suite 1259 tests, 1244 pass, the same known 14 dev-binary-override environment failures name-for-name; `pnpm test:cross-lane` 16 checks / 0 failed; no generated drift; provider contract pass; harness exit 0.

PR #355, opened 2026-08-17
fix(startup-banner): eliminate flicker, freeze, and reload-damaged intro

AlejandroTatum:
## Summary

Fixes the animated startup banner's visual problems, originally reported and verified on Ghostty:

1. **Flicker/ghosting** — the rose pulse repainted the whole header every tick (25ms) and teleported the hardware cursor on every render. Quantized the pulse step to `ROSE_PULSE_TICKS = 4` (~100ms) and render on every other tick (~20fps). Same animation duration, ~4x fewer full-header repaints.
2. **Freeze mid-flight** — pi-tui invalidates components at startup (terminal cell-size report `CSI 6;H;W t`, theme changes, resizes). The old `invalidate()` killed the ticker, freezing the logo half-written with the rose invisible. `invalidate()` is now a no-op and real teardown moved to `dispose()`.
3. **Editor region blank after /reload** — `session.reload()` re-fires `session_start` with reason `"reload"`; the intro skipped the reload guard, wrote a raw screen clear, and a second splash overlay interleaved with the reload's rendering, desyncing the TUI cursor model. The intro now returns early on `reason === "reload"`.
4. **Splash overlay instead of header** — the intro renders as a full-width, centered, non-capturing overlay on top of the chat via `ctx.ui.custom` + `overlay: true`, and hides itself once finished (`done()`).

Two commits were squashed into one (`a2e2eb9a`). No behavioral change to the rest of the extension.

## Test plan

- `node --experimental-strip-types --check extensions/startup-banner.ts` — parses clean.
- Full unit suite on the base worktree: **204/204 pass** (the same banner file content). The 16 failures in the secondary checkout are environmental (missing package-local gentle-ai binary + review mode off), unrelated to this file.
- Reproduced the original three symptoms before the fix on 2026-08-15; verified gone after applying these changes locally.

## Notes

Local copy of the fixes (never pushed before) recovered from the worktree; kept as patch + reapply script at `~/.pi/gentle-ai/` on the reporter's machine under `gentle-pi-startup-banner-fixes.patch`.

## Summary by CodeRabbit

* **Enhancements**
  * Improved startup banner animation timing for smoother performance.
  * Reload-triggered sessions now skip the introductory animation.
  * The full-screen introduction closes reliably when completed or skipped.
  * Animation continues smoothly during screen updates and cleans up correctly when dismissed.

### DIFF
--- a/extensions/startup-banner.ts
+++ b/extensions/startup-banner.ts
@@ -342,6 +342,13 @@ type LetterStroke = { orderMap: Map<string, number>; maxOrder: number };
 
 const WRITING_START_TICK = 6;
 const FALLBACK_LETTER_TICKS = 8;
+// The rose pulse advances one step every N ticks instead of every tick.
+// The pulse repaints every rose cell per step, which drives the TUI diff to
+// redraw the whole header region and teleport the hardware cursor on every
+// render. On fast terminals (ghostty) that churn shows as flicker/ghosting.
+// Quantizing the step to ~100ms keeps the gentle wave look while cutting
+// repaints by 4x after the fade-in/flash phases settle.
+const ROSE_PULSE_TICKS = 4;
 
 const LETTER_STROKES: Array<LetterStroke | null> = LETTER_SPANS.map(() => null);
 const LETTER_TICKS: number[] = LETTER_SPANS.map(() => FALLBACK_LETTER_TICKS);
@@ -605,6 +612,12 @@ export default function (pi: ExtensionAPI) {
       !process.argv.every((arg) => arg.startsWith("-") || arg.endsWith(".ts"));
     if (isCLICommand) return;
 
+    // On /reload, session.reload() re-fires session_start with reason
+    // "reload". The intro must not run there: the raw screen clear and a
+    // second overlay splash interleave with the reload's own rendering and
+    // left the editor region blank (TUI cursor model diverged from the
+    // terminal). The splash is a first-start experience only.
+    if ((_event as { reason?: string })?.reason === "reload") return;
     if (currentIntroMode() === "skip") return;
 
     // Fire-and-forget: el setup geométrico de cada letra corre en background
@@ -698,7 +711,7 @@ export default function (pi: ExtensionAPI) {
     };
 
     setTimeout(() => {
-      ctx.ui.setHeader((tui, theme) => {
+      void ctx.ui.custom((tui, theme, _keybindings, done) => {
         if (state.timer) clearInterval(state.timer);
 
         const animStart = Date.now();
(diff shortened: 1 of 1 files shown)


fedetandil:
Independent reproduction on a different terminal and a different machine — this is not Ghostty-specific.

**Environment**

- `gentle-pi` 2.5.0 (current `latest` on npm)
- `pi` 0.85.1
- Linux x86_64, Fedora 44
- Warp v0.2026.09.02, plus a bare Python `pty.fork()` with `TERM=xterm-256color`

**What I measured**

The rose is painted as `rose[255,118,195] × k`, with `k = min(1, tick/10) * pulse`, so the emitted SGR color identifies the exact frame the intro froze on. Extracting `\x1b[38;2;r;g;b m` from raw captured output and keeping the colors proportional to the rose yields the frozen tick directly.

Three consecutive runs, same machine, same version, nothing changed between them:

| run | rose color | k | frozen frame |
|-----|-----------|------|--------------|
| 1 | `(2, 1, 1)` | 0.01 | tick 0 |
| 2 | `(22, 10, 16)` | 0.09 | tick 1 |
| 3 | `(2, 1, 1)` | 0.01 | tick 0 |

It varies run to run, which is exactly the race this PR describes. A screenshot taken in Warp measured `(20,9,15)`–`(25,11,19)` — tick 1 as well. Note `k` has a `Math.max(0.01, ...)` floor, so tick 0 paints `(2,1,1)`: black on any dark background.

**On the cell-size report**

Worth noting for triage: in all three pty runs `CSI 16 t` was emitted but nothing answered it — a bare pty replies to nothing — and the intro still froze. So the cell-size report is sufficient but not necessary. The `invalidate()` broadcasts from `loadAllHighlightLanguages()` and from `themeController.applyFromSettings()` → `notifyChanged()` reach the header on their own.

One config-dependent detail that shifts the timing: with an explicit `theme` in pi's settings, `applyFromSettings()` takes the fast branch; without one it first awaits `detectTerminalBackgroundTheme({ timeoutMs: 100 })` and lands ~100ms later. That, plus how long grammar loading takes, decides which frame you get stuck on — and it is why the failure is graded rather than binary. Anyone whose invalidate lands late sees a nearly-complete rose and never reports it.

**Why this still matters**

The published `gentle-pi@latest` tarball is byte-identical to what this PR targets: `extensions/startup-banner.ts`, sha256 `037a355b...cee84d`, same `invalidate() { cleanup(); }` at the same lines. 2.3.0, 2.4.0 and 2.5.0 all shipped after this PR was opened, none of them carrying it.

The fix here matches pi's own documented contract — `examples/extensions/custom-header.ts` leaves `invalidate()` empty, and `pi-tui`'s doc comment defines it as "invalidate any cached rendering state", with real teardown belonging in `dispose()`.

Issue #367, opened 2026-08-18
bug(review): hardcoded 600s pi host-relay reviewer timeout strands large candidates in a FINALIZE retry loop


carolitascl:
## Problem

On a large high-risk review candidate, the Pi host relay (`gentle_review` FINALIZE → materialize-marked lens slots) cannot complete any lens: the locked-down reviewer `pi` subprocess is killed by a hardcoded 600s timeout, FINALIZE returns `pi-host-relay-transport-failure`, negotiated STATUS re-offers the exact same slot, and every sanctioned retry re-spends reviewer model tokens only to time out again. Three related defects make the loop worse: misleading shell guidance for `--agent pi`, a materialize-form gate that contradicts the offered slot, and no timeout knob anywhere.

## Reproduction

Environment: macOS arm64, gentle-pi 2.2.0 (global npm install), gentle-ai 2.4.0 (both the package-pinned binary and a Homebrew build), RDD mode enabled globally.

1. Have a workspace candidate large enough to be tiered high risk (observed: 180 changed files, 36,950 changed lines, 4 canonical lenses, correction budget 200; the Go-materialized lens prompt is ~1.58 MB).
2. Resolve the intended-untracked collect, run negotiated START, grant the consent envelope. START succeeds: lineage created, state `reviewing`, action `finalize`.
3. Run tests for final evidence (they pass), then call FINALIZE; acknowledge the `reviewer-model-run-forecast` with `reviewer_run_acknowledged: true`.
4. FINALIZE launches the first host-relay slot: `capture-result --agent=pi --materialize=true` (performed internally by the extension) → fresh locked-down print-mode `pi` subprocess with the 1.58 MB prompt on stdin.

## Actual behavior

**Defect A — relay reviewer timeout with no override (the blocker).**
`lib/review-host-relay.ts` pins `DEFAULT_PI_TIMEOUT_MS = 600_000`. `piTimeoutMs` is only settable through the test-injectable runner; the production FINALIZE path passes no timeout, and no environment variable or setting reaches it. Measured against the user-configured model, a single lens over this prompt class ran 478s when it succeeded and over 600s under the extension run, so the relay kills it. The typed failure says: "Re-query negotiated STATUS and relaunch only if the exact same bound slot is reoffered" — the slot is always reoffered (nothing else can satisfy it), so the sanctioned behavior is a token-burning retry loop with zero admitted results (`captured_slots: []`, `mutation_outcome: "none"` each time).

[...]

PR #371, opened 2026-08-18
fix(review): detect a linked worktree missing its project-local .pi config

### LINKED ISSUE #370: review: a linked worktree loses the project-local .pi configuration, so review fails claiming pi is an unsupported runtime
## Problem

A linked Git worktree silently loses the project-local `.pi` workspace configuration, because ignored files are not carried into a worktree. Review then fails in a way that names the runtime rather than the missing configuration.

Diagnosed by @EloyEMC on #367 and confirmed end to end: after provisioning the same project-local `.pi` configuration in the feature worktree and ensuring `GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1` was exported, the exact provider-issued `capture-result --agent=pi --materialize` flow succeeded, the locked-down Pi reviewer result was admitted, FINALIZE reached `approved`, and `review validate --gate=pre-commit` returned `allow`.

## Why it presents as "pi is not a supported runtime"

`gentleAiProcessEnvironment` (`lib/native-review-cli.ts`) already sets the handshake on every gentle-ai invocation the extension makes:

```ts
return { ...base, [GENTLE_PI_REVIEW_RELAY_CONTRACT_ENV]: GENTLE_PI_REVIEW_RELAY_CONTRACT };
```

So the extension's own calls are fine. The failure appears when the extension is not the caller: without the workspace configuration, work falls back to running the provider-issued commands by hand in a shell, where nothing exports the handshake. gentle-ai then correctly refuses `pi` and lists the three runtimes that do not need one.

Nothing in that chain mentions a worktree or a missing configuration, which is why four separate reporters on #367 landed on the same wrong conclusion about runtime support.

## Caveat on the diagnosis

The confirming run changed two things at once: it provisioned `.pi` **and** ensured the handshake export. Which of the two was load-bearing is not isolated. Both are real gaps and both are worth closing, but the report should not be read as proving that `.pi` provisioning alone is sufficient.

## Expected behavior

When Pi starts, or a workspace or worktree is registered, gentle-pi either makes its project-local configuration available in the active worktree, or detects its absence and says so plainly. Silently operating a review lineage from a worktree with no workspace integration is what turns a one-line setup gap into a runtime-support mystery.

## Test the reporter asked for, and it is the right one

Create a linked worktree, verify the host relay provisions or resolves the required workspace configuration, run `capture-result --agent pi --materialize`, and assert the result is admitted rather than refused as an unsupported runtime. Nothing today covers the worktree case.

## Related

- #367 — where this was diagnosed. Its defect A (the relay timeout) is fixed and merged in #368; this is a different defect that accumulated on the same thread.
- `Gentleman-Programming/gentle-ai#3440` — the refusal that misdirects. It names the variable now but not its value, so even an operator who reaches it cannot act without knowing `gentle-pi.review-relay/v1`.

## Credit

@EloyEMC isolated the worktree boundary and confirmed the full green path. @ElCaaarnal, @NicolasIppoliti and the original reporter each contributed an independent occurrence on published `2.4.0`, which is what made the pattern visible.

Alan-TheGentleman:
Closes #370.

A linked Git worktree never receives the project-local `.pi` configuration, because Git does not copy ignored files into a worktree. Without it the extension is not the caller, gentle-ai gets driven from a plain shell, nothing exports `GENTLE_PI_REVIEW_RELAY_CONTRACT`, and gentle-ai correctly refuses `pi` while listing the three runtimes that need no handshake. Four independent reporters on published 2.4.0 read that refusal as a statement about runtime support.

This detects the gap at the boundary that creates it, and names it.

## Scope: detection, not provisioning

@EloyEMC proposed either provisioning the configuration into the worktree or detecting its absence and reporting it. This does the second.

Provisioning means writing into a user's worktree, copying whatever their parent workspace happens to contain, and deciding what is safe to copy. It can silently duplicate stale configuration, and it is a much larger surface. Detection is small, honest, and cannot damage anything. Nothing here writes into a worktree.

## Detection condition

`inspectWorktreeWorkspaceConfig(cwd)` in `lib/worktree-workspace-config.ts`:

1. Resolve `git rev-parse --show-toplevel` and `git rev-parse --git-common-dir`, both canonicalised.
2. If `realpath(<toplevel>/.git) === commonDir`, this is the main worktree. Stop.
3. Resolve the parent workspace from the first `git worktree list --porcelain` record. A `bare` first record has no main worktree, so there is nothing to compare against and nothing is reported.
4. If that main worktree is this toplevel, this is `--separate-git-dir`, not a linked worktree. Stop.
5. Report every probed `.pi` path that exists in the parent workspace and not here.

The probed set is the project-local paths the extension actually reads: `.pi/npm/node_modules`, `.pi/settings.json`, `.pi/gentle-ai`, `.pi/agents`, `.pi/subagents`, `.pi/subagents.json`, `.pi/skills`.

Every failure to resolve is `undetermined` and reports nothing. A detector that guesses would put a second wrong explanation in front of the same user this exists to unblock.

## What a user sees

At session start, and only when something is actually missing:

```
Gentle AI: this linked Git worktree does not carry the project-local .pi configuration your parent workspace has.
Git never copies ignored files into a worktree, so nothing here is broken or corrupted. It was simply never brought across.

Worktree:         /home/you/repo-worktrees/feature
Parent workspace: /home/you/repo

Missing here, present in the parent workspace:
  /home/you/repo-worktrees/feature/.pi/npm/node_modules
    present at /home/you/repo/.pi/npm/node_modules
    project-local Pi package installs. When gentle-pi is installed here rather than globally, the extension does not load in this worktree at all
  /home/you/repo-worktrees/feature/.pi/settings.json
    present at /home/you/repo/.pi/settings.json
    project Pi settings
  /home/you/repo-worktrees/feature/.pi/gentle-ai
    present at /home/you/repo/.pi/gentle-ai
    project Gentle AI config: runtime guardrails, background-subagents policy, persona, SDD preflight

Until this is resolved, this worktree runs without that configuration.
In particular, if gentle-pi is installed project-locally, its extension does not load here, so gentle-ai ends up being driven from a plain shell.
A plain shell exports no GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1, and without that handshake gentle-ai refuses the `pi` runtime and lists claude-code, opencode and codex instead.
That refusal is correct: it names the runtime, but the cause is the missing configuration named above, not a missing runtime.

Resolve it by copying the listed paths from the parent workspace into this worktree, or by running Pi from the parent workspace.
gentle-pi will not write into your worktree on its own.
```

`/gentle:status` carries the same finding as one line, because that is the command someone already investigating "pi is not a supported runtime" reaches for:

```
Linked worktree config: 3 project-local path(s) missing that /home/you/repo has (.pi/npm/node_modules, .pi/settings.json, .pi/gentle-ai) - copy them across or run Pi from the parent workspace; until then the `pi` review runtime is refused for want of GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1
```

## What this does not touch

- The handshake requirement is unchanged. gentle-ai refusing an unhandshaken `pi` is correct, and the message says so.
- `gentleAiProcessEnvironment` is unchanged; the extension's own invocations were never the problem.
- One test asserts the message never suggests `review mode disable` or switching to another `--agent`.
- No `runtime/*.mjs` regeneration: the new module is not one of the five sources `scripts/build-git-commit-transaction-runner.mjs` generates from, and `pnpm run check:transaction-runner` confirms no drift.

## Tests

`tests/worktree-workspace-config.test.ts`, 10 tests. They create real linked worktrees rather than constructed state, including the one the reporter asked for: a linked worktree, driven through the extension's real `session_start`, asserting the missing configuration is reported there instead of surfacing later as a runtime refusal.

Revert-proof. With the linked-worktree branch in `inspectWorktreeWorkspaceConfig` reverted, 7 of the 10 go red, including the session-start and `/gentle:status` end-to-end ones. Restored, all 10 pass.

## Verification

Every CI step, foreground, all green:

- `pnpm test` — 1305 tests, 1304 pass, 0 fail, 1 skipped
- `node scripts/verify-package-files.mjs` — 163 files, 65 byte-identical v2.4.0 contract artifacts
- `pnpm run check:provider-contract` — contract 1.1.0, 8 bundle entries, 2 baselines
- `pnpm run check:transaction-runner` — 5 modules match their TypeScript sources
- `GENTLE_PI_REQUIRE_NATIVE_BINARY=1 pnpm run test:packed-runner` — gentle-pi 2.2.0, Gentle AI 2.4.0, 13 states

## Still open after this

Per @Alan-TheGentleman's routing on #367, the confirming run provisioned `.pi` and ensured the handshake export together, so which was load-bearing is not isolated. This closes the invisible half: the configuration gap now announces itself instead of arriving as a runtime-support mystery. `gentle-ai#3440` and `gentle-ai#3443` remain the threads for the refusal message itself.

## Summary by CodeRabbit

- **New Features**
  - Added detection of project-local `.pi` configuration for linked Git worktrees.
  - Session startup now warns when required configuration is missing or cannot be inspected.
  - `/gentle:status` now displays configuration status and highlights issues with warning severity.
  - Warnings identify affected capabilities and note the required review relay handshake.

- **Bug Fixes**
  - Improved visibility into incomplete worktree setup and configuration-related issues.

### DIFF
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -151,6 +151,11 @@ import {
 import type { ReviewCollectInputV3, ReviewConsentEnvelope, ReviewStatusV3 } from "../lib/review-integration-v2.ts";
 import { assertDistinctCorrectionEvidence, resolveCorrectionStep, type CorrectionEvidence, type CorrectionOutcome, type CorrectionStep } from "../lib/review-correction-lifecycle.ts";
 import { recordReviewConsentLatch } from "../lib/review-consent-latch.ts";
+import {
+	inspectWorktreeWorkspaceConfig,
+	renderWorktreeWorkspaceConfigStatusLine,
+	renderWorktreeWorkspaceConfigWarning,
+} from "../lib/worktree-workspace-config.ts";
 
 const GRAPH_V1_ORDINARY_READ_ONLY = "Graph-v1 ordinary review authority is read-only; use native compact-v2 review operations";
 import {
@@ -7415,6 +7420,21 @@ function createGentleAiExtensionForTesting(
 		} catch (error) {
 			if (ctx.hasUI) ctx.ui.notify(`Gentle AI dev binary override check failed: ${error instanceof Error ? error.message : String(error)}`, "warning");
 		}
+		// gentle-pi#370: a linked worktree starts without the ignored
+		// project-local `.pi` configuration. Say so here, at the boundary that
+		// created the gap, instead of letting it resurface much later as a
+		// gentle-ai refusal that names `pi` as an unsupported runtime.
+		try {
+			const worktreeConfig = renderWorktreeWorkspaceConfigWarning(inspectWorktreeWorkspaceConfig(ctx.cwd));
+			if (ctx.hasUI && worktreeConfig !== undefined) ctx.ui.notify(worktreeConfig, "warning");
+		} catch (error) {
+			if (ctx.hasUI) {
+				ctx.ui.notify(
+					`Gentle AI could not check this worktree's project-local .pi configuration: ${error instanceof Error ? error.message : String(error)}`,
+					"warning",
+				);
+			}
+		}
 		try {
 			const transactionRecovery = reconcileCommitTransaction(ctx.cwd);
 			if (ctx.hasUI && transactionRecovery.status !== "clean") {
@@ -7922,10 +7942,20 @@ function createGentleAiExtensionForTesting(
 			const localSddAgentOverrides = sddLocalAgentOverrideCount(ctx.cwd);
 			const modelConfig = await readModelConfigAsync(ctx.cwd);
 			const devBinary = await describeDevBinaryOverride();
+			// gentle-pi#370: whoever is already investigating "pi is not a
+			// supported runtime" reaches for this command, so the worktree
+			// configuration gap has to be visible from here too.
+			let worktreeConfigLine: string | undefined;
+			try {
+				worktreeConfigLine = renderWorktreeWorkspaceConfigStatusLine(inspectWorktreeWorkspaceConfig(ctx.cwd));
+			} catch {
+				worktreeConfigLine = undefined;
+			}
 			ctx.ui.notify(
 				[
 					"el Gentleman package is active.",
 					...(devBinary.state === "inactive" ? [] : [devBinary.line]),
+					...(worktreeConfigLine === undefined ? [] : [worktreeConfigLine]),
 					`Persona: ${readPersonaMode(ctx.cwd)}`,
 					`Global SDD agents: ${agentsInstalled ? "installed" : "not installed"}`,
 					`Global SDD chains: ${chainsInstalled ? "installed" : "not installed"}`,
@@ -7943,7 +7973,7 @@ function createGentleAiExtensionForTesting(
 					`Global model config: ${existsSync(modelConfigPath(ctx.cwd)) ? "present" : "missing"}`,
 					...describeModelConfig(ctx.cwd, modelConfig),
 				].join("\n"),
-				staleSddAssets > 0 || localSddAgentOverrides > 0 || devBinary.state !== "inactive" ? "warning" : "info",
+				staleSddAssets > 0 || localSddAgentOverrides > 0 || devBinary.state !== "inactive" || worktreeConfigLine !== undefined ? "warning" : "info",
 			);
 		},
 	});
--- a/lib/worktree-workspace-config.ts
+++ b/lib/worktree-workspace-config.ts
@@ -1,0 +1,230 @@
+// gentle-pi#370. Detects the one configuration boundary that made a setup gap
+// look like a runtime-support verdict.
+//
+// Git never copies ignored files into a linked worktree, and `.pi/` is ignored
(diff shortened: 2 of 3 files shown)


PwnLabsmx:
The detection is useful and read-only, but two boundaries still need tightening before this proves #370 closed:

- `inspectWorktreeWorkspaceConfig()` runs synchronous, unbounded Git probes from both session startup and `/gentle:status`. A stalled Git process can freeze those user-facing paths. Please make the probe asynchronous and time-bounded, with an injectable test seam; timeout and spawn failure should return `undetermined`.
- The tests instantiate `createGentleAiExtension()` directly. That cannot prove the PR's project-local-install claim: if the linked worktree lacks the package installation that loads gentle-pi, this warning code never runs. Either narrow the guarantee to sessions where gentle-pi is already globally or otherwise loaded, or prove the owning Pi startup/config-discovery boundary with an actual runtime test.

The current diff is also `+549/-1`, above the repository's 400-line review budget, so it needs a split/simplification or an explicit `size:exception`. CI is green, and the refusal correctly avoids auto-copying potentially stale configuration; these changes would keep that design while making startup bounded and the closure claim testable.

PR #458, opened 2026-08-27
fix(guard): emit guarded command permission lifecycle

### LINKED ISSUE #394: guardedCommands: emit an event before ctx.ui.confirm so notification extensions can react
## Problem

   `extensions/gentle-ai.ts` blocks the agent on `ctx.ui.confirm("Allow guarded command?", preview)` (currently around line 1109, inside
 `confirmCommand`) without emitting any event on the extension bus.

   As a result, notification extensions like `pi-smart-voice-notify` (which listens on the `pi-permission-system:permission-request` channel)
 never learn that the user is being blocked on a decision, and no sound / desktop notification / TTS fires.

   In practice: if I `Alt+Tab` away from the terminal while the agent runs, and it hits a guarded `git push` / `rm -rf` / etc., Pi sits
 silently forever waiting for me. There is no audible or visual cue that human attention is required.

   Subagent permission requests DO fire notifications, because `pi-smart-voice-notify`'s `PermissionForwardingWatcher` sees the forwarded
 permission JSON file on disk. The parent's own guarded confirms are invisible to notification extensions.

   ## Repro

   1. Install `gentle-pi` and `pi-smart-voice-notify` (with `enablePermissionNotification: true`).
   2. Ask the agent to run any guarded command in the parent session (e.g. `git push origin HEAD`).
   3. When the "Allow guarded command?" confirm appears → no notification, no sound.

   ## Root cause

   `ctx.ui.confirm` is a blocking UI primitive. It does not publish to the `pi-permission-system:permission-request` event channel that
 notification extensions monitor.

   ## Suggested fix

   Before `await ctx.ui.confirm(...)` in `confirmCommand`, publish a `waiting` event, and after it resolves publish a matching `approved` /
 `denied` event with the same `requestId`. Something like:

   ```ts
   const requestId = crypto.randomUUID();
   pi.events.emit("pi-permission-system:permission-request", {
     requestId,
     state: "waiting",
     source: "tool_call",
     message: `Guarded command awaiting confirmation: ${preview}`,
     toolName: "bash",
   });

   const approved = await ctx.ui.confirm("Allow guarded command?", preview);

   pi.events.emit("pi-permission-system:permission-request", {
     requestId,
     state: approved ? "approved" : "denied",
     source: "tool_call",
     message: `Guarded command ${approved ? "approved" : "denied"}: ${preview}`,
     toolName: "bash",
   });
 ```

Andiveli:
Closes #394

## Summary

- emit a correlated permission lifecycle around guarded command confirmations
- publish privacy-safe `waiting`, `approved`, and `denied` states through the existing extension event bus
- preserve `herdr:blocked` active/inactive compatibility for existing Herdr integrations
- preserve fail-closed behavior when the confirmation UI rejects

## Changes

| File | Change |
|------|--------|
| `extensions/gentle-ai.ts` | Emits generic permission events and Herdr compatibility events without exposing command contents. |
| `tests/gentle-ai.test.ts` | Covers ordering, correlation, approval, denial, failures, privacy, and inactive paths. |
| `tests/runtime-harness.mjs` | Models the extension event bus and verifies both lifecycle contracts end to end. |

## Test plan

- [x] `pnpm test` (1068 passed, 10 skipped, 0 failed)
- [x] `pnpm run check:runtime-modules`
- [x] Syntax checks for changed TypeScript and JavaScript files
- [x] `git diff --check`
- [x] Manually verified Herdr enters blocked state during a guarded confirmation on Linux

## Summary by CodeRabbit

- **New Features**
  - Added permission request lifecycle events for guarded command confirmations, including waiting, approved, and denied states.
  - Permission requests now use correlated identifiers while protecting command details.
  - Added blocked-command notifications when confirmation is rejected or fails.
  - Concurrent confirmations now share a consistent blocked-state lifecycle while retaining individual request tracking.

- **Bug Fixes**
  - Rejected or failed confirmations now correctly block command execution and report denial.
  - Headless and unguarded commands avoid unnecessary confirmation prompts and events.
  - Confirmation errors are handled consistently while preserving the original error.

### DIFF
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -1069,11 +1069,17 @@ async function confirmCommand(
 			toolName: "bash",
 		});
 	};
+	const emitHerdrBlocked = (active: boolean): void => {
+		events.emit("herdr:blocked", active
+			? { active: true, label: "Guarded command confirmation" }
+			: { active: false });
+	};
 
 	let approved = false;
 	let confirmationFailed = false;
 	let confirmationError: unknown;
 	emitPermissionRequest("waiting");
+	emitHerdrBlocked(true);
 	try {
 		approved = await ctx.ui.confirm("Allow guarded command?", preview);
 	} catch (error) {
@@ -1081,6 +1087,7 @@ async function confirmCommand(
 		confirmationError = error;
 	} finally {
 		emitPermissionRequest(confirmationFailed || !approved ? "denied" : "approved");
+		emitHerdrBlocked(false);
 	}
 	if (confirmationFailed) throw confirmationError;
 	if (approved) return undefined;
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -1023,10 +1023,32 @@ function evaluateSensitivePathTool(
 	};
 }
 
+type HerdrConfirmationLifecycle = {
+	begin(): void;
+	settle(): void;
+};
+
+function createHerdrConfirmationLifecycle(events: ExtensionAPI["events"]): HerdrConfirmationLifecycle {
+	let pending = 0;
+	return {
+		begin() {
+			const wasIdle = pending === 0;
+			pending += 1;
+			if (wasIdle) events.emit("herdr:blocked", { active: true, label: "Guarded command confirmation" });
+		},
+		settle() {
+			if (pending === 0) return;
+			pending -= 1;
+			if (pending === 0) events.emit("herdr:blocked", { active: false });
+		},
+	};
+}
+
 async function confirmCommand(
 	command: string,
 	ctx: ExtensionContext,
 	events: ExtensionAPI["events"],
+	herdrLifecycle: HerdrConfirmationLifecycle,
 ): Promise<ToolCallEventResult | undefined> {
 	const guardrailsConfig = loadRuntimeGuardrailsConfig(ctx.cwd);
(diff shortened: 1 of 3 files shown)


decode2 (changes requested):
Please fix the global Herdr blocked-state handling before merge.

With two overlapping guarded-command confirmations, the current flow emits `active: true` twice, then emits `active: false` as soon as the first confirmation settles, even though the second confirmation is still pending. I reproduced the sequence `[true, true, false]` at `221099e`; this can make Herdr hide an outstanding human-attention request.

Please either track the number of pending guarded confirmations and emit `active: false` only when the count reaches zero, or remove the unscoped `herdr:blocked` lifecycle from this PR. Add a concurrent-confirmation regression test for the chosen behavior. The correlated permission-request lifecycle itself looks correct.

Andiveli:
Fixed in `35fc1903`. The Herdr lifecycle now keeps an instance-scoped pending count and emits only on the `0 → 1` and `1 → 0` transitions. Two overlapping confirmations therefore keep Herdr blocked until the final confirmation settles, while the generic permission lifecycle remains independently correlated per request. Added concurrent and instance-isolation regression coverage. Full verification passes: 1068 tests, 10 skipped, 0 failed.

Alan-TheGentleman (approved):
Scoped, tested, and linked to the approved issue. The concurrency regression now keeps Herdr blocked until the final confirmation settles, and exact-head verification passed 1,077 tests with no failures. Merging.

Issue #160, opened 2026-07-16
fix(codegraph): execute npm shim safely on Windows


MarsSall:
## Bug description

The integrated Pi `codegraph` tool reports that the binary was not found on Windows even when the globally installed CodeGraph CLI is present on `PATH` and works from an interactive shell.

This is not an index problem and restarting Pi with the npm global directory on `PATH` is insufficient. The Gentle Pi extension invokes `execFile("codegraph", args)`. npm installs CodeGraph on Windows as command shims (`codegraph`, `codegraph.cmd`, and `codegraph.ps1`), not as a native executable. Node's shell-free `execFile` cannot launch those shims directly: `codegraph` returns `ENOENT`, while `codegraph.cmd` returns `EINVAL`.

Related history:

- #65 requested initial CodeGraph availability before integrated support existed.
- #106 / #108 added the constrained integrated CodeGraph tool.
- This issue is the Windows execution defect in that integrated implementation.

## Steps to reproduce

1. On Windows, install CodeGraph globally through npm.
2. Confirm `C:\Users\<user>\AppData\Roaming\npm` is present in the Pi process `PATH`.
3. Confirm `where codegraph` returns `codegraph` and `codegraph.cmd`.
4. Confirm `codegraph --help` works from Git Bash or another shell.
5. Start Pi in a real Git project root and call the integrated `codegraph` tool with `operation: init`.
6. Observe: `CodeGraph is unavailable because the codegraph binary was not found.`

Minimal Node reproduction:

```js
const { spawnSync } = require("node:child_process");

console.log(spawnSync("codegraph", ["--help"]));     // ENOENT
console.log(spawnSync("codegraph.cmd", ["--help"])); // EINVAL
```

## Expected behavior

The integrated tool should execute an npm-installed CodeGraph CLI on Windows while preserving:

- shell-free argument boundaries;
- cancellation and timeout behavior;
- current-workspace confinement;
- existing output limits and fail-closed behavior;
- no command injection through user-provided CodeGraph queries.

A safe implementation should resolve the Windows npm package entry point and invoke it with `process.execPath` plus an argument array, rather than enabling `shell: true` or concatenating a `cmd /c` command.

## Actual behavior

The shell CLI works, but the integrated Pi tool reports the binary as unavailable because `extensions/codegraph-tools.ts` uses:

```ts
execFile("codegraph", args)
```

## Environment

- OS: Windows
- Runtime: Pi with Gentle Pi CodeGraph extension
[...]

PR #553, opened 2026-09-01
fix(codegraph): resolve npm shim on Windows without shell

### LINKED ISSUE #552: codegraph tool always reports 'binary was not found' on Windows (npm shims not spawnable via execFile)
## Summary
The `codegraph` tool registered by `extensions/codegraph-tools.ts` never works on Windows: every call returns `CodeGraph is unavailable because the codegraph binary was not found`, even when `@colbymchenry/codegraph` is installed globally and its shims are on PATH.

## Root cause
`runCodeGraphCommand` (`extensions/codegraph-tools.ts:180`) runs `execFileAsync("codegraph", ...)` **without `shell: true`**. On Windows, npm only installs shims — `codegraph` (a shell script), `codegraph.cmd`, `codegraph.ps1` — and **no `codegraph.exe`**. `execFile` uses CreateProcess without a shell, which can only launch real executables, so it always fails with `ENOENT`. The tool's ENOENT fallback then reports the binary as unavailable.

Reproduced on Node v24 with `@colbymchenry/codegraph@1.2.0`:
- `execFile('codegraph')` → ENOENT
- `execFile('codegraph.cmd')` → EINVAL
- `execFile('codegraph', { shell: true })` → works ("1.2.0")

## Impact
- The codegraph-first reading rule is dead on Windows for the orchestrator **and** all subagents (same process).
- Affects every Windows user of gentle-pi 2.3.0.

## Proposed fix
Run the resolved `@colbymchenry/codegraph` package script (from its `bin` field, sibling of the `codegraph.cmd` shim in the npm global bin dir) through `process.execPath` — no shell, no command injection — falling back to `cmd.exe` resolution only as a last resort. POSIX behavior unchanged.

See PR: #<pr-number will be added>

Shammael-genios:
Fixes #552

## Problem
`runCodeGraphCommand` (`extensions/codegraph-tools.ts`) runs `execFileAsync("codegraph", ...)` without `shell: true`. On Windows, npm installs only shims (`codegraph.cmd`/`.ps1` + shell script) and no `codegraph.exe`, so plain `execFile` (CreateProcess, no shell) always fails with `ENOENT` and the tool reports `CodeGraph is unavailable because the codegraph binary was not found` on every call — for the orchestrator and all subagents.

## Fix
- Try plain `codegraph` first (POSIX and real-binary setups unchanged).
- On win32 + ENOENT: locate `codegraph.cmd` in the npm global bin dir (PATH), read the `@colbymchenry/codegraph` `bin` target from its `package.json`, and run it through `process.execPath` — **no shell, no command injection**.
- Last resort on win32: let `cmd.exe` resolve the shim (`shell: true`).
- New exported helpers `findCodeGraphCmdOnPath` / `codeGraphNodeScript` with a platform-neutral unit test.

## Validation
- `node --experimental-strip-types --test tests/codegraph-tools.test.ts`: new test passes; the only failing test (symlink EPERM on Windows) is pre-existing and also fails on main.
- Full suite vs main: **+1 test, +2 pass, -1 fail** (remaining failures are pre-existing Windows-environment issues; suite is green on the Linux/macOS CI).
- Verified end-to-end on Windows: `codegraph query` returns real results in the orchestrator and in subagents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows compatibility when running CodeGraph commands.
  * Added fallback handling for npm-installed command shims.
  * Ensured queries containing special shell characters are passed safely as literal arguments.
  * Improved handling of outdated, invalid, or unavailable command paths.

* **Tests**
  * Added coverage for locating Windows command shims, resolving their underlying scripts, preserving quoted paths, and safely handling special characters.


dnlrsls (dismissed):
Requesting changes on the Windows fallback at `extensions/codegraph-tools.ts:263`. `args` includes user-controlled query text and the workspace path, and `shell: true` hands those values to `cmd.exe`. A bounded Windows probe confirmed that metacharacters in a query execute a second command. This is a merge-blocking command-injection regression, and it conflicts with the tool's no-shell contract.

Please remove the shell fallback entirely. After the initial Windows `ENOENT`, walk the trimmed and unquoted PATH entries, resolve each shim's adjacent `@colbymchenry/codegraph` package bin, and invoke it with `execFile(process.execPath, [script, ...args], runOptions)`. Preserve the existing cwd, `AbortSignal`, `maxBuffer`, stdout/stderr, and error behavior, and rethrow the original `ENOENT` if no safe entry point resolves.

Please also add regression coverage for metacharacter-bearing argv and a later valid shim after an invalid one. In `tests/codegraph-tools.test.ts:244`, restore an originally absent PATH by deleting `process.env.PATH` instead of setting it to an empty string. The focused suite currently passes 10/10, so it does not detect this vulnerability. Please link the approved root issue #160 rather than relying on unapproved duplicate #552.

PR #743, opened 2026-09-08
feat(extensions): add /history interactive bash session browser

sdiazbarraza:
# Proposal: /history — Interactive Bash History Browser

## Why

Pi users currently have no way to recall a bash command from earlier in the same session without scrolling through transcripts. Power users bring Bash `Ctrl+R` muscle memory into Pi but find nothing equivalent. A slash command that browses the current session's bash history fills that gap and shortens "rerun that 3-minute build step" from minutes to seconds.

## What

- Two commands register: `/history` (primary) and `/r` (2-character alias).
- `/history` enumerates every bash command executed in the current Pi session, sourced from both direct `!cmd` invocations and the LLM-driven `bash` tool calls.
- Presentation order: newest first, deduplicated by exact command string.
- The picker shows a single-select list with an inline typing filter. The title rendered to the user is `🔍 [gentle-pi] Selecciona un comando:`.
- Selection behavior: the chosen command is injected into the active editor via `ctx.ui.setEditorText(command)`. The user can submit or further edit before sending.
- Empty-history behavior: when no bash commands exist yet in this session, the extension posts a non-blocking `ctx.ui.notify("No hay comandos de bash en el historial de esta sesión.", "info")` and exits without rendering the picker.
- Cancellation behavior: pressing `Esc` (or `Ctrl+C`) inside the picker exits silently — the editor input is left untouched and no exception is raised.

## User stories

- As a Pi user, I want `/history` to fuzzy-find any bash command I ran earlier (direct `!cmd` or LLM-driven bash) so I can re-execute it without scrolling.
- As a Pi user, I want `/r` to be a 2-character alias for `/history`.
- As a Pi user, an empty history shows a non-blocking notification, not an error.
- As a Pi user, pressing `Esc` in the picker leaves the editor untouched.

## Non-goals

---

Closes or tracks: #742 in `Gentleman-Programming/gentle-pi`.

### What this PR adds

- A new extension file `extensions/history-search.ts` (307 LOC) that registers two slash commands under `gentle-pi`:
  - `/history` — `Buscar en el historial de comandos de terminal de la sesión`.
  - `/r` — `Alias rápido para /history`.
- A fuzzy-filtered interactive picker over the current session's bash history (both `BashExecutionMessage` entries and assistant `toolCall` blocks where `name === 'bash'`).
- Selection injects the command into the editor via `ctx.ui.setEditorText(...)`.
[...]

PR #747, opened 2026-09-08
feat(extensions): add /input interactive user prompt recall

sdiazbarraza:
### Problem

Pi users cannot recall a long prompt they sent to the LLM earlier in the session without scrolling through the transcript. Bash `Ctrl+R` provides exactly this muscle memory: incremental fuzzy search over what the user has typed before. The `/history` slash command in #743 covers bash executions only; user prompts remain unsearchable.

This PR adds `/input` (alias `/i`) — a sibling slash command that emulates Bash `Ctrl+R` over the user's own prompts in the current session.

### What this PR adds

- New extension file `extensions/user-prompt-recall.ts` (~120 LOC) registering two slash commands:
  - `/input` — `Buscar en mis prompts enviados al LLM durante la sesión`.
  - `/i` — `Alias rápido para /input`.
- New shared helper `lib/fuzzy-recall-picker.ts` (~180 LOC) extracted from the existing `/history` picker so both commands render the same TUI.
- Source: `ctx.sessionManager.getEntries()` filtered to `message.role === "user"`. String content is taken directly; multipart content joins its `text` blocks and skips entries with no text.
- Selection injects the chosen prompt into the editor via `ctx.ui.setEditorText(...)`.
- Empty-prompt-history → `notify`. Esc → silent exit. Outside TUI mode → `notify`.
- Four new test files under `tests/user-prompt-recall.*.test.ts` (~37 new tests).
- Modification: `extensions/history-search.ts` loses 190 LOC of duplicated picker plumbing and re-imports the helper from `lib/`. The 30 existing history-search tests stay green without modification.

### Verification

- Full repo suite: 1713 pass / 0 fail / 10 pre-existing skips.
- 30 history-search tests untouched and still green.
- New user-prompt-recall tests cover: command registration, prompt extraction across string/array/mixed-content cases, picker render contents, selection → exactly-one `setEditorText`, Esc → zero `setEditorText`, dedupe of identical prompts, mode guards.
- `git diff main -- lib/fuzzy-recall-picker.ts extensions/user-prompt-recall.ts` is the entirety of the new surface area.

### Diff stat

```
 extensions/history-search.ts        | -190 lines (extraction)
 extensions/user-prompt-recall.ts    | +120 lines (new)
 lib/fuzzy-recall-picker.ts          | +180 lines (new helper)
 tests/user-prompt-recall.*.test.ts  | +330 lines (new tests)
```

Total net: ~440 LOC new (within the 400-line review budget once the history-search extraction is netted against the new content; well-bounded PR).

### Behavior compatibility

- `/history` semantics unchanged. It still shows bash executed. The two commands are siblings, not replacements.
- No new npm dependencies. Reuses `@earendil-works/pi-coding-agent` (DynamicBorder) and `@earendil-works/pi-tui` (SelectList, Container, Key, matchesKey).
- Zero `as unknown as` / `as never` / `@ts-ignore` casts in the new files.
- No edits to `lib/native-choice-list.ts` or `lib/native-fullscreen-interaction.ts`.

### Why a lib helper now (not before)

The earlier `/history` proposal explicitly rejected promoting the picker to `lib/` because there was a single consumer. `/input` adds a second consumer with identical TUI requirements. The rejection no longer applies.

### Local smoke test

```bash
git clone https://github.com/sdiazbarraza/gentle-pi.git -b feat/user-prompt-recall
cd gentle-pi && pnpm install && pnpm test    # 1713 pass / 0 fail
```

### DIFF
--- a/extensions/history-search.ts
+++ b/extensions/history-search.ts
@@ -2,16 +2,13 @@ import type {
 	ExtensionAPI,
 	ExtensionCommandContext,
 } from "@earendil-works/pi-coding-agent";
-import { DynamicBorder } from "@earendil-works/pi-coding-agent";
 import {
-	Container,
-	Key,
-	matchesKey,
-	SelectList,
-	Text,
-	type SelectItem,
-	type SelectListTheme,
-} from "@earendil-works/pi-tui";
+	createFuzzyRecallPicker,
+	dedupeNewestFirst,
+	filterByQuery,
+	scoreMatch,
+	type PickerTheme,
+} from "../lib/fuzzy-recall-picker.ts";
 
 const EMPTY_HISTORY_MESSAGE =
 	"No hay comandos de bash en el historial de esta sesión.";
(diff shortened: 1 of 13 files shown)


sdiazbarraza:
Tracking: sibling PR to #743 (`/history`). Same repo policy applies — draft until the maintainer reviews the design and gives a directional signal.

Pickup: `5d6735ca` on top of `d87082d2`. Diff vs. main: `+`1186 / `-`190. The extraction in `extensions/history-search.ts` removes the duplicated picker plumbing and re-imports it from `lib/fuzzy-recall-picker.ts`. The new picker helper has two consumers (`/history` and `/input`) so the `lib/` move is justified.

cc @maintainers — happy to split this into chained PRs (extract → /input) if preferred over a single draft.

Issue #644, opened 2026-09-06
bug(extensions): pi-pretty hard-crashes extension load when @heyhuynhgiabuu/pi-pretty is not installed


carolitascl:
### Summary

`extensions/pi-pretty.ts` calls `requireFromRealPackage("@heyhuynhgiabuu/pi-pretty")` at module top level with no error handling. In a source checkout where dependencies were never installed (no `node_modules` at all), this turns a missing cosmetic dependency into a hard extension-load failure on every Pi startup, with a confusing `Cannot find module` error that points at `package.json`.

### Error

```text
Error: Failed to load extension "<local gentle-pi checkout>/extensions/pi-pretty.ts": Failed to load extension: Cannot find module '@heyhuynhgiabuu/pi-pretty'
Require stack:
- <local gentle-pi checkout>/package.json
```

### Root cause

Top-level, unguarded require in `extensions/pi-pretty.ts`:

```ts
const packageJsonPath = realpathSync(
  fileURLToPath(new URL("../package.json", import.meta.url)),
);
const requireFromRealPackage = createRequire(packageJsonPath);
const piPrettyModule = requireFromRealPackage("@heyhuynhgiabuu/pi-pretty");
```

pi-pretty is a purely cosmetic enhancement (syntax highlighting, tool-output formatting). When the package cannot be resolved, the exception propagates out of the module and Pi's loader reports the extension as failed, instead of gentle-pi skipping the enhancement with an actionable message.

Observed state of the checkout where this fired:

- `package.json` declares `"@heyhuynhgiabuu/pi-pretty": "0.6.14"` in `dependencies` — the pin is correct
- `pnpm-lock.yaml` is in sync (lockfile entries present for 0.6.14)
- The npm package and the pinned 0.6.14 tarball are intact on the registry (verified via `npm view`)
- `node_modules/` does not exist at all in the checkout — dependencies were simply never installed there, and neither README nor a CONTRIBUTING file documents an install step for running extensions from source

So this is not a registry or versioning problem: any source checkout used directly as a Pi extension directory without `pnpm install` first will crash the pi-pretty extension load on every startup.

### Steps to reproduce

1. Fresh clone of gentle-pi (no `pnpm install`, so no `node_modules`)
2. Point Pi at the checkout's `extensions/` directory
3. Start Pi
4. The pi-pretty extension fails to load with `Cannot find module '@heyhuynhgiabuu/pi-pretty'`

### Impact

[...]

Issue #238, opened 2026-07-30
pi-pretty never loads in compiled Pi binary (createRequire intercepted by jiti)


XaviCode1000:
### Summary

**pi-pretty never works for compiled Pi binary users.** The root cause is not pi-pretty\s dependencies per se — it\s that the Pi compiled binary (Bun) intercepts `createRequire` calls through jiti\s virtual module resolver, so even `requireFromRealPackage("@heyhuynhgiabuu/pi-pretty")` cannot resolve a package that is physically installed in `node_modules`.

Related issue: #60 (same symptom, Windows user after Pi v0.80.2).

---

### Root Cause

Gentle-pi\s `extensions/pi-pretty.ts` does:
```js
const requireFromRealPackage = createRequire(packageJsonPath);
const piPrettyModule = requireFromRealPackage("@heyhuynhgiabuu/pi-pretty");
```

`createRequire` is supposed to use **native filesystem resolution**. Under the compiled Pi binary, jiti intercepts it and routes everything through its **virtual module resolver**, which only knows about `@earendil-works/pi-ai` (root), `/oauth`, and `compat`. The bare specifier `@heyhuynhgiabuu/pi-pretty` cannot be found even though it IS installed at:

`~/.pi/agent/npm/node_modules/@heyhuynhgiabuu/pi-pretty/`

Confirmed: running the same `require.resolve(...)` on Node.js v24 resolves correctly. Under the Pi compiled binary (v0.82.1 and v0.83.0 tested), it fails with `ResolveMessage: Cannot find module`.

### Additional issue: pi-pretty v0.6.16+

Starting from v0.6.16, pi-pretty added `@earendil-works/pi-coding-agent` as a peer dependency and uses a **top-level synchronous** `require("@earendil-works/pi-coding-agent")` in its dist. Since pi-coding-agent is only available as a Pi runtime virtual module (not installed on disk), this also fails.

Gentle-pi pins `"@heyhuynhgiabuu/pi-pretty": "0.6.14"` in `package.json`, but Pi\s root lockfile has `"^0.6.14"` which resolves to the latest 0.6.x (currently 0.6.21). Version 0.6.14 uses **dynamic `await import()` inside a try/catch** for pi-coding-agent, which is safe. Version 0.6.16+ uses a synchronous `require()` at the top level with no fallback.

### Impact

- The pi-pretty extension **never loads** for anyone using the compiled Pi binary (the normal distribution channel)
[...]

PR #835, opened 2026-09-10
fix(extensions): load pi-pretty via ESM import under compiled Pi (#238)

MzaGuille:
## Summary
Fixes hard startup failure under the **compiled Pi (Bun) binary** when loading `extensions/pi-pretty.ts`:

```text
ResolveMessage: Cannot find module '@heyhuynhgiabuu/pi-pretty' from '.../gentle-pi/package.json'
```

Even when `@heyhuynhgiabuu/pi-pretty@0.6.14` is installed and Node `createRequire(gentle-pi/package.json)` resolves it.

Closes / addresses **#238**. Softens **#644** (missing package → graceful skip instead of extension-load crash).

## Root cause (community + Omarchy)
`realpathSync` + `createRequire` under compiled Pi ignores package `main`/`exports` (`pi-pretty` uses `main: dist/index.js`). ESM import resolves correctly. Reinstall / nesting / bumping to 2.5.0 does **not** fix — the wrapper was unchanged.

## Change
- `extensions/pi-pretty.ts`: cached `await import("@heyhuynhgiabuu/pi-pretty")` inside the async factory; keep quiet-tools / `PRETTY_DISABLE_TOOLS` behavior; try/catch skip if unresolved.
- `tests/package-manifest.test.ts`: assert ESM import, reject `createRequire`/`realpathSync`.

## Test plan
- [ ] `pnpm test` / package-manifest tests for the wrapper assertions
- [ ] Smoke on compiled Pi binary with gentle-pi + pi-pretty installed (no `ResolveMessage`)
- [ ] Smoke with pi-pretty absent → warning, Pi still starts
- [ ] Confirm `pi -ne` still works

## Notes
Consensus on #238 preferred this minimal async-import over bundling (packaging risk).
Omarchy temporary mitigations while waiting: `pi -ne` or settings `extensions` filter excluding `extensions/pi-pretty.ts`.


MzaGuille:
The wrapper in `extensions/pi-pretty.ts` is the right fix for #238. The try/catch skip also covers #644.

Please do not merge as-is: `tests/package-manifest.test.ts` was replaced with `PLACEHOLDER_LOAD_FROM_FILE`, which deletes the rest of the suite (~1553 lines). Restore that file from `main` and change only the existing pi-pretty assertion so it requires `import("@heyhuynhgiabuu/pi-pretty")` and rejects `createRequire(` / `realpathSync(`.

Verified on Omarchy with pi 0.85.1 compiled binary: ESM import loads pi-pretty, and `pi --offline -p` starts without the ResolveMessage.

Alan-TheGentleman:
Thank you for isolating and validating the compiled-runtime loader failure. #1108 has now merged the ESM loader fix together with the pi-pretty upgrade and byte-exact tool-result regression coverage, so this PR is superseded.

PR #842, opened 2026-09-10
fix(startup): keep banner animation alive across pi-tui invalidation

### LINKED ISSUE #836: bug(startup): banner animation is frozen — early invalidation leaves only the first frame
### Before submitting

### Problem

Starting `pi` with no arguments renders the startup banner rose, but it stays static: the animation never plays and the banner freezes on its first frame.

Reading the current implementation (`extensions/startup-banner.ts`, main at commit 9ec677f): during `session_start` the extension installs the banner header via `ctx.ui.setHeader` and starts `setInterval(... tui.requestRender(), 25)` to animate it. The component handed back to Pi returns `invalidate() { cleanup(); }`, and `cleanup()` cancels both the interval timer and its listeners. An early invalidation during startup can therefore cancel the animation while only the first frame has been drawn, leaving the banner visibly frozen for the rest of the session.

Related but distinct: #744 reports a nearly black first frame with a missing wordmark under a specific remote terminal; this report covers the general frozen-animation behavior in an ordinary local terminal.

### Steps to reproduce

1. Install gentle-pi 2.5.0 (main at commit 9ec677f) with the startup banner enabled, running on Pi 0.85.1.
2. Run `pi` with no arguments in an interactive terminal.
3. Watch the startup banner rose after it appears.
4. Observe that it never animates; it remains on its first frame.

Expected:
The startup banner plays its animation frames normally. If Pi invalidates the banner component early, the lifecycle should still leave a sensible rendered state (for example the completed animation frame) instead of a frozen first frame.

Actual:
The rose appears but remains static; no animation is observed. This is consistent with an early `invalidate()` during startup cancelling the 25 ms render interval through `cleanup()` while the first frame stays on screen.

Please fix this in a way compatible with the current implementation of the banner extension — not by reapplying any previous local workaround — and add coverage for the component lifecycle/invalidation path so an early invalidation during startup cannot freeze the banner.

### gentle-pi version

2.5.0 (main at commit 9ec677f)

### Pi version

0.85.1

### Operating system

Linux

```text
No error output is produced; the banner simply never animates.
```

emanuelcasco:
## Summary

Root cause: pi-tui calls `invalidate()` on every component when the async cell-size response lands right after `ui.start()`. The banner used that hook to `cleanup()`, so the 25ms timer died at tick 0.

`dispose()` is the hook Pi uses for header replacement. `session_shutdown` cleanup is unchanged.

## Linked issue

Closes #836

## PR type

- [x] Bug fix (`type:bug`)

## Summary

- Stop treating pi-tui `invalidate()` as banner teardown.
- Keep the animation timer alive across the first-boot cell-size invalidation.
- Tear down only on `dispose()` and `session_shutdown`.

## Changes

| File | Change |
| --- | --- |
| `extensions/startup-banner.ts` | Move cleanup from `invalidate()` to `dispose()`. |
| `tests/startup-banner.test.ts` | Match the header lifecycle type to `dispose()`. |

## Test plan

- [x] `pnpm exec node --experimental-strip-types --test tests/startup-banner.test.ts`
- [x] `git diff --check`
- [x] Manual: start `pi` with no args and confirm the rose and wordmark animate instead of freezing on frame 0.

## Contributor checklist

- [x] Linked an approved issue.
- [x] Selected exactly one PR type.
- [x] No shell scripts changed; shellcheck is not applicable.
- [x] No documentation change is required.
- [x] Conventional commit used with no `Co-Authored-By` trailers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup banner lifecycle handling so animations continue updating correctly and stop cleanly when the banner is disposed or the session ends.
  * Ensured cleanup actions are consistently performed during banner shutdown.

### DIFF
--- a/extensions/startup-banner.ts
+++ b/extensions/startup-banner.ts
@@ -1051,7 +1051,7 @@ export default function (pi: ExtensionAPI) {
 
             return out;
           },
-          invalidate() {
+          dispose() {
             cleanup();
           },
         };
--- a/extensions/startup-banner.ts
+++ b/extensions/startup-banner.ts
@@ -1051,6 +1051,7 @@ export default function (pi: ExtensionAPI) {
 
             return out;
           },
+          invalidate() {},
           dispose() {
             cleanup();
           },
--- a/tests/startup-banner.test.ts
+++ b/tests/startup-banner.test.ts
@@ -18,7 +18,48 @@ test("startup branch lookup uses direct git argv and hides its Windows child", a
 		command: "git",
 		args: ["-C", "/repo with spaces & metacharacters", "branch", "--show-current"],
 		options: { encoding: "utf8", shell: false, windowsHide: true },
-	}]);
+    }]);
+});
+
+test("startup banner keeps animating after invalidate and cleans up on dispose", async (t) => {
+	t.mock.timers.enable({ apis: ["setTimeout", "setInterval", "Date"] });
+	t.mock.method(fs, "readFile", async () => JSON.stringify({ showRose: true, showTextLogo: true, color: "pink" }));
+	syncBuiltinESMExports();
+	t.after(() => { t.mock.restoreAll(); syncBuiltinESMExports(); });
+	const argv = process.argv;
+	process.argv = ["node"];
+	t.after(() => { process.argv = argv; });
+	for (const [key, value] of [["rows", 40], ["columns", 160]] as const) {
+		const descriptor = Object.getOwnPropertyDescriptor(process.stdout, key);
+		Object.defineProperty(process.stdout, key, { configurable: true, writable: true, value });
+		t.after(() => descriptor ? Object.defineProperty(process.stdout, key, descriptor) : Reflect.deleteProperty(process.stdout, key));
+	}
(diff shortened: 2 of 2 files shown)


emanuelcasco:
Added the lifecycle regression for #836.

`invalidate()` is now a no-op (pi-tui layout-cache), and the test asserts the animation timer keeps firing after it and stops on `dispose()` / `session_shutdown`.

barbatdev:
Thanks Emanuel, checked this locally and the lifecycle change makes sense. Pi uses `invalidate()` for render invalidation and calls `dispose()` when replacing the header.

We tested the regression both ways: your new test fails without the fix at “invalidate() must not stop the animation timer”, then passes with it restored. All 6 banner tests pass, and a separate Cortex QA pass found no new issue attributable to this change.

Looks good within this scope. Thanks for adding the regression test!

barbatdev (approved):
Approved for the scoped lifecycle fix. Verified the regression fails without the production change and passes with it restored (6/6 focused banner tests). Review details are in the preceding comment. Repository CI still needs verification before merge.

PR #912, opened 2026-09-11
fix(agents): record why a subagent run ended when its tool call is aborted

### LINKED ISSUE #911: bug(agents): an aborted subagent tool call leaves no result, no record, and no termination reason
### Problem

`subagent_run` and `subagent_continue` in `task` mode wait for the child inside the tool call. Pi hands every tool an `AbortSignal` as the third `execute()` argument, but the subagent tools registered `execute(_id, params, _signal, _onUpdate, ctx)` and dropped it.

When the caller's turn is aborted — a human interrupting a running turn, or a host-side timeout — the tool call therefore ends **with no result**, the child process **keeps running**, and the task record gets **no termination reason**. Nothing in the transcript or in the durable record explains why the call disappeared, so the run is unattributable after the fact.

### Steps to reproduce

1. Start a task-mode subagent (`subagent_run` with `mode: "task"`) whose child does not finish immediately.
2. While that tool call is in flight, abort the caller's turn (interrupt in the TUI, or abort the signal the host passes to `execute()`).
3. Observe: no tool result is delivered for that call, the child is still alive, and the task store holds no terminal record explaining the outcome.

Relevant code: `extensions/gentle-agents.ts` registers its tools through a local `tool()` helper whose wrapper ignored the signal; `lib/agents-runner.ts` already supports `cancel(id)`, which persists a terminal record. Other tools in this package (for example `extensions/codegraph-tools.ts`) do declare and honour `signal: AbortSignal | undefined`.

**Expected:** aborting the caller's signal cancels the child through the runner, so the normal stop lifecycle runs and the record is persisted as `cancelled`; the tool call resolves with that outcome instead of vanishing; and the user is told the call was aborted, including the abort reason when the host supplies one.

**Actual:** the tool call ends silently, the child keeps running to completion unobserved, and no reason is recorded anywhere.

Related: #906 covered a task reaching a terminal state without its record being persisted; this is the different case where the *caller* goes away and nothing records a reason at all. #791 covers the `/reload` cancellation path.

Alan-TheGentleman:
## What

Thread the `AbortSignal` Pi hands to a tool's `execute()` into the subagent launch path, and use it to cancel the child through the runner when the caller goes away.

Closes #911.

## Why

The subagent tools registered `execute(_id, params, _signal, _onUpdate, ctx)` and dropped the signal. A task-mode call aborted by the host (human interrupt, timeout) therefore ended **with no result**, left the child **running**, and recorded **no termination reason** — the run was unattributable after the fact.

## How

- `extensions/gentle-agents.ts` — the local `tool()` helper now forwards the signal; `launch()` accepts it. On abort it calls `runner.cancel(task.id)`, so the normal stop lifecycle runs and the record is persisted as `cancelled`, then notifies the user with the abort reason. The listener is removed in a `finally`, so a normal completion cannot trigger a second cancellation. `run` and `continue` pass their signal through.
- `lib/agents-runner.ts` — adds `abortReasonText()` to render an `Error`/string reason (empty when there is none) so the message says *why*, not just *aborted*.
- Tests — an integration test drives the registered `subagent_run` with a real `AbortController` and asserts the recorded status is `cancelled`, the tool call resolves instead of vanishing, the warning names the abort, and the child was terminated; plus unit coverage for `abortReasonText`. The test harness's `Registered.execute` widens `signal` from `undefined` to `AbortSignal | undefined`.

## Evidence

- `pnpm test`: 2204 tests, 2195 pass, 0 fail (9 pre-existing skips).
- `node scripts/check-types.mjs`: `205 recorded diagnostic(s), no regressions`.
- Diff: 4 files, +79/−17.

## Review

Reviewed under the native review gate (`review-6bc942344f9c385b`, tier high, 4 lenses; the reliability lens run failed and was reconciled by native status, which declared the candidate approved). Authority acknowledged and burned; delivery is ordinary repository policy.

## Summary by CodeRabbit

- **New Features**
  - Added cancellation support for running agent tasks when an abort request is received.
  - Users are notified when a task is cancelled, including the cancellation reason when available.
  - Cancelled tasks now complete their lifecycle and retain their records.

- **Bug Fixes**
  - Ensured cancelled child processes terminate cleanly instead of continuing to run.

### DIFF
--- a/extensions/gentle-agents.ts
+++ b/extensions/gentle-agents.ts
@@ -11,7 +11,7 @@ import { sidebarPart } from "../lib/shell-sidebar.ts";
 import { invalidateSidebar } from "../lib/shell-sidebar-layout.ts";
 import { AGENT_MODE, discoverAgents, loadAgentsConfig, resolveAgentProfile, type AgentDefinition, type AgentMode } from "../lib/agents-config.ts";
 import { isFinished, TASK_STATUS, TaskStore, type AskRequest, type TaskRecord } from "../lib/agents-protocol.ts";
-import { AgentRunner, piCommand, type AskAnswer, type RunnerDeps, type SddChangeSelection, type TaskRequest } from "../lib/agents-runner.ts";
+import { AgentRunner, piCommand, abortReasonText, type AskAnswer, type RunnerDeps, type SddChangeSelection, type TaskRequest } from "../lib/agents-runner.ts";
 import { ChildMessenger, type IpcEndpoint } from "../lib/agents-messaging.ts";
 import { hasReviewSessionPermission, resolveCanonicalGitRepositoryIdentitySync, type ReviewSessionManager } from "../lib/review-session-standing-permission.ts";
 import { historyDir, loadStoredTask, pruneHistory, saveTask } from "../lib/agents-history.ts";
@@ -685,7 +685,7 @@ export default function gentleAgents(pi: ExtensionAPI, env: NodeJS.ProcessEnv = 
 		};
 	};
 
-	const launch = async (ctx: ExtensionContext, request: TaskRequest): Promise<ToolText> => {
+	const launch = async (ctx: ExtensionContext, request: TaskRequest, signal?: AbortSignal): Promise<ToolText> => {
 		// Bounded live observation only. Native send owns the fresh policy decision;
 		// child execution never starts a telemetry policy process or renewal timer.
 		const owner = metricsOwner;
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman:
CI note: `review-repository-windows` failed once on `candidate ownership is private and durable before worktree add` (`candidate view owner preparation failed`, `lib/review-candidate-view.ts`). That file is untouched by this diff, and the job passed on rerun — a Windows worktree flake, not a regression from this change. All three checks are green now: verify, review-repository-windows, CodeRabbit.

PR #913, opened 2026-09-11
fix(agents): bound background completion delivery to the current turn

### LINKED ISSUE #867: bug(subagents): stale completion results replay after compaction
### Before submitting

### Problem

Completed background subagent results can be delivered much later as new conversation messages, after their work has already been consumed and the related pull requests have been merged. This creates noise, makes completed work appear active again, and forces the orchestrator to repeatedly classify stale results.

The observed messages were old `Subagent ... finished` notifications for tasks completed significantly earlier. They arrived sequentially after conversation compaction/resumption rather than when each task settled.

### Steps to reproduce

1. Start multiple background subagent tasks in a long Pi session.
2. Consume their results and continue working until the tasks and related work are complete.
3. Let the conversation undergo context compaction or resume the session afterward.
4. Observe the conversation for additional subagent completion messages.
5. Notice that already-consumed task results may reappear one by one as if they were newly completed.

Expected:
- Each subagent completion notification is delivered at most once and promptly enough to represent the current task state.
- Results already consumed by the parent are not replayed after compaction, reload, or session resumption.
- If delayed delivery is unavoidable, the message is explicitly marked stale and does not re-enter the normal user-message flow.

Actual:
- Old completion results are emitted later as new conversation messages.
- Several already-finished tasks can be replayed sequentially after their corresponding work has been merged.
- The parent must repeatedly explain that each notification is stale and requires no action.

### gentle-pi version

2.5.0

### Pi version

0.85.1

### Operating system

macOS

```text
Subagent gentle-ai-verify (...) finished.
Subagent gentle-ai-worker (...) finished.

These notifications arrived after their results had already been consumed and the associated work completed.
```

Alan-TheGentleman:
## Linked Issue

Closes #867

---

## Summary

Background subagent completions were handed to the host as `deliverAs: "followUp"` messages. The host documents that mode as "waits for agent to finish, delivered only when agent has no more tool calls", so the notification's lifetime was unbounded: in an orchestrator that chains tool-calling turns, it was delivered whenever the whole run finally went idle.

Measured in a real session transcript, this is not a hypothesis:

| task | parent pulled the result | notification delivered | delay |
|---|---|---|---|
| `mtx6xx54-1-kwyn` | 17:02:39 | 18:00:52 | **58 min** |
| `mtx7lz4k-2-hnhg` | 17:19:40 | 18:09:45 | **50 min** |

There was no compaction entry anywhere in that session file, and the notification was flushed immediately after an assistant message that had no tool calls — the next idle turn boundary. The parent had already consumed both results through the `result` tool and reported both PRs before the notifications arrived.

Gentle Pi now owns completion delivery:

- **Idle parent** → delivered immediately, exactly as before, so the wake-up behaviour is unchanged.
- **Busy parent** → held and flushed at the next turn boundary, instead of whenever the host queue happens to drain.
- **Already consumed** (the parent pulled it with `result`, or resumed it with `continue`) → dropped silently; a consumed result is never announced on top of itself.
- **Stale** (settled ≥90s before the flush) → delivered as durable transcript-only content through `appendEntry` + a registered entry renderer. It never re-enters the model context, and it names the task and how long ago it settled.
- **Owned by another session** → dropped. Completion delivery was previously the only archiving path without the session gate that notifications and queries already had, so a task launched in one session could deliver its own environment into another.

The delivery mode on the flush path is `steer`, not `followUp`, and this is the crux of the fix: while the parent streams, steering is polled every turn and injected before the next LLM call, so the delay is bounded to the current turn. `followUp`'s drain point is the run loop's stop branch, which is exactly the unbounded lifetime this bug was about.

---

## Changes

| File / Area | What Changed |
|-------------|-------------|
| `lib/agents-completion-delivery.ts` (new) | Dependency-free, clock-injected completion queue: `enqueue` once per task id, `consume` that also remembers unknown ids, `takeDeliverable(now)` with the explicit `stale` decision, `dropAll`. Exports `STALE_COMPLETION_MS = 90_000` with the justification in-file |
| `extensions/gentle-agents.ts` | Completion queue wiring: `settleCompletion` at `onFinish` (visibility conditions untouched), `flushCompletions` on `turn_end` with `agent_end`/`agent_settled` as safety nets, immediate flush when no agent run is active, `deliver` gated on the owning session and delivered as `steer`, stale entries appended with a renderer, `consume` at the `result` tool, task-mode `waitFor`, and `continue`, and `dropAll` on `session_start`/`session_shutdown` |
| `tests/agents-completion-delivery.test.ts` (new) | 10 unit tests: settle order, duplicate-enqueue no-ops, consumed drop, unknown-id consume memory, idempotent consume, both sides of the stale boundary |
| `tests/gentle-agents.test.ts` | 5 integration tests (hold-until-turn-end and delivered exactly once, stale goes to `appendEntry` only, consumed dropped at flush, session restart never replays, cross-session drop) plus the two option pins updated to `steer`, each with a comment naming the drain behaviour it protects against. The fake-`pi` harness gained `registerEntryRenderer`/`entryRenderers` |

---

## Test Plan

```bash
node --experimental-strip-types --test tests/agents-completion-delivery.test.ts tests/gentle-agents.test.ts
pnpm test
pnpm run typecheck
pnpm run check:runtime-modules
```

Observed:

- Focused: `tests 77 / pass 77 / fail 0`
- `pnpm test`: exit 0 — `tests 2217 / pass 2208 / fail 0 / skipped 9` (all pre-existing Windows-only skips), provider contract mirror check passed, runtime harness exit 0
- `pnpm run typecheck`: `types: 205 recorded diagnostic(s), no regressions`
- `pnpm run check:runtime-modules`: `runtime matches TypeScript sources`

**RED-first evidence.** With the tests pinned to the new behaviour and the pre-change `extensions/gentle-agents.ts`, exactly 5 integration tests fail, each with `1 !== 0` on the recorded sends — for example `a background completion settling while the parent agent runs is delivered exactly once at the next turn end` → `'nothing enters the conversation while the parent agent run is active'` (`tests/gentle-agents.test.ts:1761`). Pre-change code sent the completion while the parent was mid-run: that failure *is* the defect.

**Falsification of the mode specifically.** Reverting only the one mode token to `followUp` fails exactly 2 tests with `{ + deliverAs: 'followUp', - deliverAs: 'steer', triggerTurn: true }`; restoring it returns the file to a byte-identical hash (`3528e30a…`), with final occurrence counts `steer=1` (completions) and `followUp=2` (notification/query, intentionally untouched).

An independent read-only verification pass reproduced the falsification in its own copy, re-audited at-most-once, the no-replay-after-resume paths, and the session gate, and left the worktree byte-identical (hashes captured before and after).

---

## AI Assistance

Material assistance was used. Pi coding agent (el Gentleman harness) with delegated Pi subagents:

- a read-only exploration pass that mapped the delivery path and found the missing session gate on completions;
- a bounded implementation writer that produced the change under strict TDD and reported its own falsification evidence;
- an independent read-only verifier that reproduced the falsification, re-audited the invariants, and **rejected the first implementation** on the delivery mode, with a host-source trace (`agent-session.js` / `agent.js` / `agent-loop.js`) showing that `turn_end` still runs inside an active run and that `followUp` is drained only at the run's stop branch.

The human reviewed the diff, the design forks, and the final delivery mode, and authorized the issue approval and the dependency install in the fresh worktree. Full diff reviewed by the human before committing.

---

## Notes for Reviewers

Three things worth your attention, in order of how much I would challenge them:

1. **`steer` is a deliberate behaviour change.** A fresh completion now enters the context before the parent's next LLM call rather than after the whole run. That is requirement #1 of the issue ("promptly enough to represent the current task state") and it is what makes the delay bounded, but it does mean a completion can interleave between tool calls of the orchestrating run. `onNotification` and `onQuery` keep `followUp` untouched.
2. **`STALE_COMPLETION_MS = 90_000` makes a long parent turn transcript-only.** A single turn legitimately exceeding 90s means its held completion becomes TUI-only instead of a conversation message. That is requirement #3, and the result stays reachable through `subagent_status`/`subagent_result`, but it is a real threshold decision and I would rather it be argued about than assume it.
3. **`agent_end` is kept as a flush trigger** even though the host may still auto-retry or continue after it, because a run can end without another `turn_end` (abort, or a run with no assistant message). The host compensates for an early post-run return via `hasQueuedMessages()` + `continue()` (`agent-session.js:776-778, 808-810`), and steering stays bounded there; `agent_settled` is documented in-code as a normally-no-op safety net.

Test-harness limitation to be aware of: the fake `pi` records `sendMessage`/`appendEntry` synchronously and has no `isStreaming` model, so the suite cannot observe host-side injection timing. The mode pins are therefore the regression guard — `tests/gentle-agents.test.ts:1275-1280` and `:1766-1772`. A future edit that silently restores `followUp` will fail them.

Deliberately not changed: the `onNotification`/`onQuery` paths, and the display-only `/new` + `/resume` behaviour pinned at `tests/gentle-agents.test.ts:1682-1702`.

Known residual risks, stated rather than smoothed over: a crash between enqueue and flush loses a completion (equivalent to the old behaviour, where a held `followUp` also died with the process); if the extension is registered mid-run, one completion can still take the host queue for that run and self-correct afterwards; and the stale card's live TUI rendering was verified through the renderer contract and the installed type definitions, not in a real interactive session.

Also observed while doing this work, and **not** caused by this change: `gentle_review_capture_group` reported `prepared_reviewers: 4, submitted_reviewers: 3` with `native-operation-failed` on the last lens, twice in a row across two different repositories, and bound STATUS then reconciled the lineage to `approved`. Either the engine tolerates a missing lens or the relay bookkeeping is wrong; it deserves its own issue.

## Summary by CodeRabbit

* **Improvements**
  * Background agent completions are delivered at the next turn boundary while the parent agent is still running.
  * Completions are delivered in settlement order and only once.

* **Bug Fixes**
  * Older completions are marked as stale and shown in the transcript without being added to model context.
  * Pending completions are cleared when sessions restart or shut down, preventing results from appearing in the wrong session.

### DIFF
--- a/extensions/gentle-agents.ts
+++ b/extensions/gentle-agents.ts
@@ -9,6 +9,7 @@ import { keyHint, type ExtensionAPI, type ExtensionContext } from "@earendil-wor
 import { Text, type TUI } from "@earendil-works/pi-tui";
 import { sidebarPart } from "../lib/shell-sidebar.ts";
 import { invalidateSidebar } from "../lib/shell-sidebar-layout.ts";
+import { createCompletionQueue } from "../lib/agents-completion-delivery.ts";
 import { AGENT_MODE, discoverAgents, loadAgentsConfig, resolveAgentProfile, type AgentDefinition, type AgentMode } from "../lib/agents-config.ts";
 import { isFinished, TASK_STATUS, TaskStore, type AskRequest, type TaskRecord } from "../lib/agents-protocol.ts";
 import { AgentRunner, piCommand, type AskAnswer, type RunnerDeps, type SddChangeSelection, type TaskRequest } from "../lib/agents-runner.ts";
@@ -37,6 +38,7 @@ export const AGENTS_WIDGET_KEY = "gentle-agents";
 export const AGENTS_COMMAND_NAME = "gentle:agents";
 export const AGENTS_RESULT_TYPE = "gentle-agents.result";
 export const AGENTS_MESSAGE_TYPE = "gentle-agents.message";
+export const AGENTS_STALE_RESULT_TYPE = "gentle-agents.stale-result";
 const COLLAPSE_KEY_DEFAULT = "ctrl+shift+a";
 const VIEW_KEY_DEFAULT = "alt+a";
 const STOP_KEY_DEFAULT = "alt+s";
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman:
Merge update: `main` moved with #912 (abort handling), which changed the same call site as this fix — the task-mode wait that now also marks a completion consumed. Resolved in `e99f4479` by keeping both behaviours: the consume sits inside the abort listener's `try` block, next to the runner wait, so an aborted run still records its termination and a completed one is still marked consumed. Both new test blocks were kept (git had closed the last completion test and the new abort test with a single brace).

Validation after the merge: focused trio 129/129, `pnpm test` 0 failures, `typecheck` no new diagnostics, `check:runtime-modules` clean.

The merge resolution changed the reviewed bytes, so the native review was re-run on the merged candidate rather than carried over: lineage `review-771076b18a225c7e` (risk high, 4 lenses, 98 changed lines), approved and acknowledged. The original lineage for this branch was `review-48d6c35ecbbc4c5a`.

One thing worth flagging for the maintainers: across both review runs here, and one on a different repository today, `gentle_review_capture_group` reported `prepared_reviewers: 4, submitted_reviewers: 3` with `native-operation-failed` on the *last* lens (order 3), deterministically, and bound STATUS then reconciled the lineage to `approved`. Either the engine tolerates a missing lens or the relay bookkeeping is wrong. It is unrelated to this PR but deserves its own issue.

Issue #941, opened 2026-09-12
bug(review): select-intended-untracked rejects provider-issued selection binding, blocking START for fully-tracked candidates


jmasenjowork-glitch:
### Before submitting

- [x] I searched open and closed issues and did not find a report of this problem.
- [x] I reviewed this report and removed credentials, tokens, private paths, hostnames, and other sensitive data.

### Problem

Native ordinary review cannot START when the candidate has zero untracked files but the repo has a large unrelated untracked inventory (here: 426 files - QA screenshots plus MCP-server logs; candidate = 3 tracked modified files, 352 changed lines). `gentle_review {operation: select-intended-untracked}` with the provider-issued `selectionBinding` and the only honest value (`intendedUntracked: []`, matching the provider's own projection `intended_untracked: []` + precomputed proof) is rejected every time. Effect: the RDD gate blocks delivery with no valid forward route; the ledger/apply side is complete and green (73/73 self-tests).

### Steps to reproduce

1. Repo with committed SDD change artifacts + large unrelated untracked inventory (inventory digest `sha256:a4ea…`, stable across calls - no filesystem churn).
2. Apply a slice touching only tracked files (candidate target `sha256:540a…`).
3. `gentle_review {operation: inspect}` -> `action: start`, `next_transition: {kind: collect, reason_code: intended_untracked_selection_required}`, projection paths = 3 tracked files, `intended_untracked: []`.
4. `gentle_review {operation: select-intended-untracked, selectionBinding: <exact provider-issued bytes>, intendedUntracked: []}` -> rejected (see logs). Retried across fresh inspects with machine-verified bytes (canonical CLI oracle: eligible list byte-identical to `git status` untracked sorted list, 426 entries; escape level and short fields cross-checked).

### Expected and actual behavior

Expected: the pre-lineage selection is accepted (empty selection is consistent with the provider's own projection), unblocking START - as the non-empty selection did for a prior lineage whose candidate genuinely contained untracked files.
Actual: deterministic rejection. No lineage is created, no authority burned, no mutation performed in any attempt.

### gentle-pi version

2.5.0

### Pi version

0.85.1

### Operating system

Linux

### Relevant logs or error output (optional)

```text
attempts 1-4: {"operation":"select-intended-untracked","status":"blocked",
  "outcome":"intended-untracked-selection-binding-rejected",
  "mutation_performed":false,"mutation_outcome":"none"}
attempt 5 (machine-built bytes, round-trip verified):
[...]

PR #942, opened 2026-09-12
fix(review): name the check that refused an intended-untracked selection

L4XB:
Refs #941. **This does not fix the rejection the reporter hit — it makes the rejection say which check failed**, which is what the report is actually stuck on. Details and evidence below, including what I could rule out.

## What #941 is stuck on

`select-intended-untracked` decides eight conditions in one boolean and answers every one of them with the same string:

```ts
const rejected = input === undefined || canonicalReviewCaptureBinding(input) !== canonicalBinding
  || exactCollectArgument(input, "target_identity") !== status.targetIdentity
  || exactCollectArgument(input, "projection") !== status.projection.projection
  || exactCollectArgument(input, "base_tree") !== status.projection.baseTree
  || exactCollectArgument(input, "candidate_tree") !== status.projection.currentCandidateTree
  || !Array.isArray(eligible) || selected.reason !== undefined
  || selected.intendedUntracked!.some((path) => !eligible.includes(path));
if (rejected) return { …, outcome: "intended-untracked-selection-binding-rejected", … };
```

A caller holding provider-issued bytes cannot tell a status that moved under it (retry after a fresh inspect) from a path outside the eligible list (fix the selection) from an unreadable `eligible_paths_json` (a provider problem). The report shows the consequence: five attempts of the same call with machine-verified bytes, and no forward route.

After this change each condition names itself, in the same order, with the same `outcome` kept for compatibility:

| reason | fired by | `field` |
| --- | --- | --- |
| `selection-input-absent` | the status carries no matching collect input | — |
| `binding-mismatch` | the submitted bytes are not this input's canonical binding | — |
| `status-field-mismatch` | the live status no longer matches the bound value | the field that moved |
| `eligible-paths-unreadable` | `eligible_paths_json` absent, duplicated, or not JSON | — |
| `untracked-selection-invalid` | `validateNativeStartUntrackedSelection` refused the pair | — |
| `path-not-eligible` | a selected path is not in the eligible list | the refused path |

`nativeStartRejection` has carried `reason`/`field` for START since it was introduced, so this is the shape the file already uses.

On the one judgement call worth a reviewer's eye: `path-not-eligible` echoes back a path **the caller itself just submitted**, in the response to that same call — no information the caller did not already have, and with a 400-entry inventory it is the only way to tell which entry was refused. Dropping the `field` there is a one-line change if you would rather it never appear.

## What I could rule out

The empty-selection path itself is sound. I reproduced the reporter's shape in the existing harness — candidate touching only tracked files, an unrelated untracked inventory (including a path with spaces), `intendedUntracked: []` — and it **starts**, taking the `exclude` scope with `intended_untracked: []`:

```
starts[0].intendedUntrackedSelection = { argumentTokens: […],
  value: '{"schema":"gentle-ai.review-intended-untracked-selection/v1","untracked_scope":"exclude","expected_untracked_inventory":"sha256:aaa…","intended_untracked":[]}' }
```

That is now a test, so the case cannot regress. It also means the rejection in #941 comes from one of the identity/binding/eligible-list terms against the native provider's payload, which I cannot reproduce from here — hence the reason field rather than a guess at a fix. Once this lands, @jmasenjowork-glitch's next attempt reports which one, and that is a one-line answer instead of a five-attempt investigation.

Things I checked and found innocent: `validateNativeStartUntrackedSelection` accepts `{scope: exclude, inventory, intendedUntracked: []}` (`extensions/gentle-ai.ts:5009` only rejects a non-empty selection under `exclude`); `exactCollectArgument` does no tokenizing, so a path containing spaces is not truncated; `isCanonicalProcessString` imposes no length bound, so a 426-entry `eligible_paths_json` is not rejected for size anywhere in the TypeScript layer.

## Verification

- `node --experimental-strip-types --test tests/review-controller-native-routing.test.ts` → **68 pass, 0 fail** (was 65; three new cases).
- Negative control — reverting only `extensions/gentle-ai.ts` and keeping the tests → 4 fail: both existing `pre-lineage intended-untracked selection …` cases (the per-case `reason` assertions) and both moved-status cases. The empty-selection test passes in both states, which is what makes it a regression guard rather than a restatement of the change.
- `node scripts/check-types.mjs` → `203 recorded diagnostic(s), no regressions`. The gate also reports one pre-existing improvement in `tests/review-candidate-view.test.ts TS2540`, which is already true on a clean `origin/main` (verified by stashing), so `scripts/types-baseline.json` is deliberately left untouched here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for intended untracked-file selections, preventing selections from proceeding when their status, bindings, or eligible paths no longer match.
  - Rejected selections now include a specific reason and, when applicable, the affected field.
  - Added safeguards to prevent work from starting when live status data has changed.

- **Improvements**
  - Empty selections are now accepted when no untracked files are involved, allowing the operation to proceed with an exclusion submission.

### DIFF
--- a/extensions/gentle-ai.ts
+++ b/extensions/gentle-ai.ts
@@ -5015,6 +5015,49 @@ function validateNativeStartUntrackedSelection(value: Record<string, unknown>): 
 	};
 }
 
+// gentle-pi#941: the pre-lineage selection had eight ways to be refused and one
+// opaque outcome for all of them, so a caller holding provider-issued bytes
+// could not tell a moved status from an ineligible path -- and the only route
+// left was to send the same call again. START rejections have carried
+// `reason`/`field` since they were introduced (nativeStartRejection above);
+// name the failing check here the same way. The checks keep their original
+// order, so which one is reported is the first that fails, exactly as the
+// short-circuiting condition decided before.
+const INTENDED_UNTRACKED_BINDING_FIELDS = ["target_identity", "projection", "base_tree", "candidate_tree"] as const;
+
+function intendedUntrackedSelectionRejection(
+	input: ReviewCollectInputV3 | undefined,
+	canonicalBinding: string,
+	status: ReviewStatusV3,
+	eligible: unknown,
+	selected: NativeStartUntrackedSelection,
+): { readonly reason: string; readonly field?: string } | undefined {
+	if (input === undefined) return { reason: "selection-input-absent" };
+	if (canonicalReviewCaptureBinding(input) !== canonicalBinding) return { reason: "binding-mismatch" };
+	const bound: Record<(typeof INTENDED_UNTRACKED_BINDING_FIELDS)[number], string> = {
+		target_identity: status.targetIdentity,
+		projection: status.projection.projection,
+		base_tree: status.projection.baseTree,
+		candidate_tree: status.projection.currentCandidateTree,
+	};
+	// A field that no longer matches the live status means the binding was
+	// issued against a different target, projection or tree -- the selection is
+	// stale rather than wrong, and a fresh inspect is the forward route.
+	for (const field of INTENDED_UNTRACKED_BINDING_FIELDS) {
+		if (exactCollectArgument(input, field) !== bound[field]) return { reason: "status-field-mismatch", field };
+	}
+	// The eligible list is the provider's own `eligible_paths_json`; unreadable
+	// here means it was absent, duplicated, or not valid JSON in the status.
+	if (!Array.isArray(eligible)) return { reason: "eligible-paths-unreadable" };
+	if (selected.reason !== undefined) return { reason: selected.reason };
+	// Echoes back a path the caller itself submitted -- no information the
+	// caller did not already have, and with a large inventory it is the only
+	// way to tell which entry of a long selection was refused.
+	const ineligible = selected.intendedUntracked!.find((path) => !eligible.includes(path));
+	if (ineligible !== undefined) return { reason: "path-not-eligible", field: ineligible };
+	return undefined;
+}
+
 function nativeStartRejection(reason: string, field?: string): Record<string, unknown> {
 	return {
 		operation: REVIEW_CONTROLLER_OPERATION.START,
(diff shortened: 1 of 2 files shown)


LaurenteEber:
There is a concrete committed-range variant of this rejection that can inform the follow-up fix. This is not evidence that #941 has the same cause, and it does not require expanding this diagnostic-only PR.

### Observed on published gentle-pi 2.5.0 (macOS)

The tracked worktree was clean, with an existing committed candidate and an unchanged unrelated untracked inventory. An ordinary START with an explicit base commit and `committedOnly: true` returned the expected non-empty base-diff and an intended-untracked selection binding. Immediately submitting that exact returned binding with `intendedUntracked: []` and the same explicit worktree returned `intended-untracked-selection-binding-rejected`, with `mutation_performed: false` and `mutation_outcome: none`. The default workspace inspection instead saw an empty tracked diff. No code changes, authority reset/recovery, or binding rewriting were used between those calls.

### Installed-source diagnosis

In `extensions/gentle-ai.ts` from 2.5.0:

- START resolves the base to a canonical commit and passes `baseRef` plus `committedOnly: true` to the initial status negotiation (approximately lines 6015–6044).
- `SELECT_INTENDED_UNTRACKED` negotiates status again with `cwd` and optional signal only (5978–5981). It does not retain the committed-range selectors from the START that emitted the binding.
- The exact-binding/identity checks then compare the original range binding against the default workspace status (5984–5988). The aggregate rejection does not identify which individual condition fired, but the selector omission is directly visible in the installed source.
- There is a second loss at the recursive START (5990–5991): its input carries only ordinary mode and the untracked selection fields. It also drops `baseRef`, `committedOnly`, `policyPath`, and `focus`. Fixing only the status recheck would therefore leave the continuation incomplete. Policy/focus loss is a static finding, not the cause demonstrated by this runtime occurrence.

### Suggested follow-up boundary

Keep the rejection checks. Retain the validated pre-lineage route in the host, bound to the exact provider-issued binding, session, and canonical worktree. Preserve the already-resolved base commit, committed-only semantics, and original policy/focus through both the fresh status check and the subsequent START. Do not make the caller reconstruct selectors or relax provider binding validation. Unknown/stale routes and conflicting contexts for the same binding should fail closed.

The regression worth adding is the composition: clean tracked workspace + committed-range START + unrelated untracked inventory + selection `[]`, asserting that both the recheck and final START retain the canonical base and committed-only flag. Existing workspace-selection tests around `tests/review-controller-native-routing.test.ts:913–951` do not exercise that combination. Also preserve negative controls for inventory/candidate drift, conflicting roots, and replay, and test policy/focus transport separately.

Evidence boundary: the facade rejection was observed in the live Pi host; the causal path was then checked statically in the installed package. No new regression test, patched build, or end-to-end fix verification has been executed. This comment was prepared with AI assistance and is posted with the user's authorization; private repository details and candidate identifiers are omitted.

L4XB:
Thanks — this is the kind of report that makes a diagnostic worth having, and both source claims hold on current `main`, not only in the published 2.5.0:

- `extensions/gentle-ai.ts:7356` — the SELECT re-negotiation passes `{ cwd: defaultCwd, ...signal }` and nothing else. The `baseRef` / `committedOnly` selectors the START used to issue the binding are not carried.
- `extensions/gentle-ai.ts:7368` — the recursive START's `input` is `{ mode: ORDINARY, untrackedScope, expectedUntrackedInventory, intendedUntracked }`. `baseRef`, `committedOnly`, `policyPath` and `focus` are all dropped, so fixing only the status recheck would leave the continuation reading a different range than the one the caller asked for.

I agree this does not belong in this PR: it is a behaviour change, and this one only names the check that already refused.

**What it does do for your case is turn the aggregate into a discriminator.** The order in `intendedUntrackedSelectionRejection` is deliberate, so the reason you get back tells you which of two things happened:

- `selection-input-absent` — the default-workspace status emits no intended-untracked collect input at all. That is the reading where the empty tracked diff means there was nothing to select against, and the binding never got as far as being compared.
- `status-field-mismatch` with `field: "base_tree"` (or `candidate_tree`) — the input is there, and the named field is the one that moved. That is the reading where the committed-range binding is being compared against a default-workspace projection, which is the selector loss at 7356 stated in one word.

`binding-mismatch` would mean something else again: the input is present and its own capture binding differs, which points at the binding rather than at the range.

I cannot reproduce your exact state here (it needs the committed candidate and the untracked inventory you had), so I am not claiming which of the two fires for you. But re-running the same two calls against this branch answers it in one line, and whichever it is narrows the follow-up: `selection-input-absent` says the fix has to reach the status emission, `status-field-mismatch` says carrying the selectors at 7356 is enough for the rejection and 7368 is the separate half.

On the boundary you suggest: keeping the rejection checks and retaining the validated pre-lineage route bound to the exact provider-issued binding sounds right to me, and the policy/focus loss at 7368 is worth its own line in that follow-up even though, as you say, this occurrence does not demonstrate it.

PR #980, opened 2026-09-13
fix(banner): count the MCP servers the session loads

### LINKED ISSUE #979: Startup banner MCP stat counts configured servers, not enabled ones
### Before submitting

### Problem

The startup banner stats (the side status bar shown at session start, with rows like `GIT:`, `PATH:`, `MCP:`) reports `MCP: N server(s)` where N counts every server **configured** in the global `~/.pi/agent/mcp.json` — including servers that are **disabled** and therefore never connect or register any tools.

`extensions/startup-banner.ts` reads the raw config file and counts keys (session_start handler, around lines 656–664):

```ts
const raw = await readFile(
  join(os.homedir(), ".pi", "agent", "mcp.json"),
  "utf8",
);
const cfg = JSON.parse(raw);
mcpServersCount = Object.keys(cfg.mcpServers || {}).length;
```

That count is rendered as `MCP: ${mcpServersCount} server(s)` in all three stat layouts (wide row, narrow rows, and the `narrowRows` array — lines ~863, ~890–891, ~911).

With pi-mcp-adapter, a server can be disabled (`"disabled": true` in any MCP config, or `/mcp disable <server>`, which persists the flag in the project-local `.pi/mcp.json`); disabled servers make no connections, no authentication, and register no tools — they only stay visible in config and status. The banner still counts them, so the panel overstates the MCP surface the session actually has available. Layered config also escapes the count: servers defined only in a project-layer MCP config are invisible to the global-file parse.

### Steps to reproduce

1. Configure two MCP servers in `~/.pi/agent/mcp.json` and add `"disabled": true` to one of them (or run `/mcp disable <server>` in a project).
2. Start a new pi session with gentle-pi loaded.
3. The stats panel shows `MCP: 2 server(s)`, while the `/mcp` panel reports 1 enabled server (1 disabled).

Expected:

`MCP:` reflects the **enabled** servers the session actually loads. pi-mcp-adapter already publishes a read-only, machine-readable status snapshot (`MCP_STATUS_EVENT` / `McpStatusSnapshot`) with per-server `status`/`disabled` plus `connectedCount` and `disabledCount`; the banner could consume that instead of parsing the raw config file.

Actual:

`MCP:` reports every key configured in the global `~/.pi/agent/mcp.json`, including disabled servers and ignoring project-layer MCP config.

### gentle-pi version

2.5.0 (running from source, commit 6e4478c0)

### Pi version

0.85.1

### Operating system

macOS

_No response_

L4XB:
Fixes #979

## The problem

The banner read the global config file and reported its key count:

```ts
const cfg = JSON.parse(raw);
mcpServersCount = Object.keys(cfg.mcpServers || {}).length;
```

That is not the number of servers the session has. It is wrong in two directions at once:

- a server carrying `"disabled": true` connects to nothing, authenticates nothing and registers no tools — and still counted;
- a server configured only in the **project** layer never appeared, because only `~/.pi/agent/mcp.json` was parsed.

There was a third, quieter one: the `catch` set the count to `0`. A single unreadable file discarded everything, rather than the layer it could not read.

## The fix

`countEnabledMcpServers(cwd, read?)` reads the layers lowest-precedence first, lets a later layer replace an earlier entry of the same name, and counts what is left that is not disabled. So `/mcp disable <server>` in a project turns a globally configured server off in the banner too, which is what the `/mcp` panel already shows.

The count is a pure function of the two file bodies with the reader injected — the same shape `readGitBranch` already uses for `execFile` — so the tests state the contract without touching a filesystem.

## Tests

Four cases in `tests/startup-banner.test.ts`, each verified against a mutant:

| mutation | result |
| --- | --- |
| ignore the `disabled` flag again | 2 fail |
| read only the global layer | 1 fail |
| let an unreadable layer abort the whole count | 3 fail |
| stop a later layer overriding an earlier one | 1 fail |

The layer helper asserts the two config paths really are distinct before each case, so a collapsed path list fails loudly instead of measuring one file twice.

Full suite, same checkout with and without the change: **16 failures both ways**, identical sets (`sdd-selection-transport`, review-status and symlink candidate-view tests) — pre-existing here. Tests go 2360 → 2364, the four added ones passing.

## One thing I could not verify

The issue suggests consuming pi-mcp-adapter's `MCP_STATUS_EVENT` / `McpStatusSnapshot`, which would be the better source. That adapter is not a dependency of this package and nothing in the tree references those symbols, so I could not build against them. This reads config instead, which keeps the change self-contained.

For the same reason the project-layer path `<cwd>/.pi/mcp.json` follows the issue's description rather than something I could check here. It is a single array in `mcpConfigPaths()` if it needs correcting, and the tests address the layers through that function rather than hard-coding paths.

## Summary by CodeRabbit

- **Bug Fixes**
  - Startup banners now show an accurate count of enabled MCP servers.
  - Servers configured across global and project-level settings are included, with higher-priority settings taking precedence.
  - Disabled servers and unavailable, unreadable, or malformed configuration layers are handled safely.
  - MCP configurations using supported alternate naming conventions are recognized.

- **Tests**
  - Added coverage for layered configuration, precedence rules, disabled servers, shared settings, alternate configuration formats, and missing files.


L4XB:
Both findings taken at `7d27fc9`, and both hold up. I did not take the summary on trust though, because neither `loadMcpConfig` nor `toServerEntries` exists in this repo, so I fetched `pi-mcp-adapter@2.34.0` and read its `config.ts`.

**The layers.** `getConfigSources` builds six unconditional sources in this order, and `loadMcpConfig` merges them lowest to highest:

```
~/.config/mcp/mcp.json        GENERIC_GLOBAL_CONFIG_PATH
~/.agents/mcp.json            AGENTS_GLOBAL_CONFIG_PATHS[0]
~/.agents/mcp/mcp.json        AGENTS_GLOBAL_CONFIG_PATHS[1]
<PI_AGENT_DIR>/mcp.json       getPiGlobalConfigPath()
<cwd>/.mcp.json               getProjectConfigPath()
<cwd>/.pi/mcp.json            getProjectPiConfigPath()
```

So the order you listed is right. `mcpConfigPaths` now returns all six.

Four more adapter sources exist and I am deliberately not mirroring them, with a comment saying so rather than leaving the gap silent: exclusive-config mode, opt-in host discovery, ancestor discovery, and the package / agent-plugin / Claude-plugin configs. None of those can be resolved from a config path alone, and walking them would make the banner a second implementation of the loader instead of a reading of it. If the maintainer would rather have exactness there, the issue's own suggestion is better than either: consume the adapter's `McpStatusSnapshot` and stop parsing files.

**The entries.** `toServerEntries` keeps an entry only when `isRecord` accepts it, `isRecord` being `typeof value === "object" && value !== null && !Array.isArray(value)`. So a null, a primitive or an array never reaches the session, and `servers.set(name, entry ?? {})` was counting exactly those as enabled. Fixed with the same predicate.

One more thing fell out of reading `validateConfig`: it reads `raw.mcpServers ?? raw["mcp-servers"]`, so a config using the hyphenated spelling was being skipped entirely. That is now read too, with a cell.

`isServerDisabled` in the adapter's `types.ts` is `definition?.disabled === true`, which is what this already used, so that half needed no change.

Tests go from 8 to 12. The layer helper is keyed by layer name rather than destructuring the first two paths, and the invalid-entry table now expects zero where it expected one. Five mutations, all killed:

```
killed  only the two Pi-owned layers are read again
killed  a non-object entry is counted again
killed  the mcp-servers alias is dropped
killed  the layer order is reversed
killed  disabled is ignored
```

`pnpm run typecheck`: 200 recorded diagnostics, no regressions.

L4XB:
On the outside-diff note about `servers.set(name, entry)` (`extensions/startup-banner.ts`): you were right, and my code comment was wrong. I checked `pi-mcp-adapter` 2.34.0 again. `mergeServerMaps` merges field by field, with `merged = { ...base, ...next }` in `mergeBuiltInAgentPluginEntries`. It drops only transport and url-bound auth fields when those change, never `disabled`, and `isServerDisabled` is `definition?.disabled === true`. So a `disabled: true` from a lower layer stays in force under a higher layer that omits it.

Fixed in 970155d7:
- The banner merges entries the same way: `servers.set(name, { ...servers.get(name), ...entry })`.
- The comment now describes the merge instead of claiming a replacement.

Two new tests:
- `MCP stat keeps a disabled flag that a higher layer does not override` counts 0. It fails on the previous commit.
- `MCP stat lets a higher layer turn a server back on with disabled: false` counts 1. That is the explicit override `/mcp enable` writes when a lower layer disables a server.

All 8 MCP tests pass. The 5 artwork and animation tests in `startup-banner.test.ts` fail in this environment both before and after the change. `check-types.mjs` reports no new diagnostics.

Issue #995, opened 2026-09-13
feat(shell): unify responsive status across footer and sidebar


AutanaSoft:
### Before submitting

These first-person statements remain unchecked until the user explicitly affirms them.

- [x] I searched open and closed issues and did not find a request for this feature.
- [x] I reviewed this request and removed credentials, tokens, private paths, hostnames, and other
      sensitive data.

### Problem or opportunity

Gentle Pi presents related runtime information in the responsive shell footer and the fullscreen
Status sidebar, but each surface currently handles width constraints and visual hierarchy as a
separate concern.

On narrow terminals, the footer can drop execution metadata or integration status. In fullscreen
mode, the Status card can consume more vertical space than necessary. The effective RDD state must
also remain visible without being inferred from opaque integration strings.

Users would benefit from one coherent shell-status experience for project, model, context, usage,
review, and integrations across terminal widths and TUI modes.

### Proposed outcome

Deliver the unified experience as three independent, reviewable work units:

1. Preserve execution metadata and integration status in responsive footer layouts, including safe
   truncation for Unicode branch names.
2. Expose effective structured RDD state as `ON`, `OFF`, or unknown.
3. Compact the fullscreen Status sidebar:
   - show the project basename instead of its complete path;
   - keep branch and dirty count on the same logical line;
   - preserve the session name;
   - render Model before Review in responsive columns;
   - render Context before Usage in responsive columns;
   - omit only the provider name from sidebar usage;
   - preserve all windows from the first usage limit;
   - keep Integrations full-width and opaque;
   - stack column groups when the available width is insufficient.

The traditional bottom bar must preserve its semantics. Every rendered line must fit the supplied
visible width, including ANSI sequences and Unicode characters.

The work should be delivered through three separate PRs in this order:

```text
responsive shell footer
  -> structured RDD status
  -> compact Status sidebar
```

The integration branch demonstrates compatibility only and should not become a monolithic PR.

### One combined PR

Rejected because the integrated comparison contains 1,260 changed lines across 11 files. That review
load would obscure the boundaries between behaviors and make regressions harder to isolate.

### Completely independent changes

Keeping the branches independent protects review size, but without shared validation it does not
show that their overlapping shell and Status changes work together.

### Infer RDD from integrations

Rejected because integration statuses are opaque strings. RDD must come from its structured source.

The recommended approach keeps three work units and uses an integrated branch only for end-to-end
validation.

### Additional context

Related issues:

[...]

Issue #713, opened 2026-09-08
bug(shell): preserve context gauge with long Windows paths


Fivoryu:
### Before submitting
- [x] I searched open and closed issues and did not find a report of this problem.
- [x] I reviewed this report and removed credentials, tokens, private paths, hostnames, and other sensitive data.

### Problem
Gentle Shell's status bar can hide the context gauge when the current working directory is a long Windows path. The bar drops the `ctx` segment while retaining less critical trailing information, making context usage unavailable at a glance.

### Steps to reproduce
1. Use gentle-pi's Gentle Shell on Windows with Pi 0.85.1 or newer.
2. Start Pi from a repository whose working directory has several nested path segments.
3. Observe the status bar at a normal terminal width.

### Expected and actual behavior
Expected: The bar compacts the Windows path and keeps the context gauge visible, for example `ctx ▱▱▱▱▱▱▱▱ 0%`.

Actual: The full Windows path consumes the available width, and the bar removes the context gauge or truncates it before the user can read it. The same compaction logic also fails to shorten Windows paths because it only recognizes `/` separators.

### gentle-pi version
2.5.0 (git main)

### Pi version
0.85.1

### Operating system
Windows

### Relevant logs or error output (optional)
```text
_No response_
```

PR #1002, opened 2026-09-14
fix(shell): make the status footer responsive

### LINKED ISSUE #715: Shell bar drops extension statuses on narrow terminals (wrap instead of pop) + visibleWidth under-counts wide emoji
## Summary

On narrow terminals (e.g. half-screen layouts, ~100 cols), the Gentle Shell bar silently drops **all** extension statuses (`ctx.ui.setStatus` entries: memory/MCP/LSP/…) because `renderShellBar` resolves overflow by popping trailing segments — and statuses are appended last. A second, related bug makes lines that *were* measured as fitting physically clip at the screen edge: `visibleWidth` counts wide emoji as 1 column while terminals render them as 2.

## Environment

- gentle-pi 2.5.0, pi 0.85.1, Linux (Ghostty)
- Terminal width ~100 cols (half screen)
- Six extension statuses active: engram, FFF, MCP (`pi-mcp-adapter`), LSP (`pi-lens`), ponytail, session name

## Observed behavior (PTY captures, escalating width)

| cols | statuses visible |
|------|------------------|
| 220  | all six |
| 160  | through MCP/LSP |
| 120  | through the second status |
| 100  | first status partially clipped mid-word |
| 80   | none (bar itself truncated) |

At 100 cols the first status was cut mid-text (`🧠 tmp · re…`-style clipping) even though the measurement said it fit — that's the emoji under-count: the wrap decision used a width that did not match what the terminal actually rendered.

## Root causes

1. `lib/shell-bar.ts` → `renderShellBar` overflow loop pops segments from the end:
   ```ts
   while (segments.length > 1 && visibleWidth(left) > width) {
       segments.pop();
       left = joinSegments(segments, theme);
   }
   ```
   Statuses are the trailing segments, so they die first. The comment above `compactModel` says "the trailing statuses … survive on ordinary terminal widths", but on half-screen widths the base bar alone (brand + location + model + ctx + cost) already consumes ~70 cols.

2. `visibleWidth` returns 1 for wide emoji (`🧠` U+1F9E0, `🔌` U+1F50C, `🐴` U+1F434 are all East-Asian-Width Wide/Fullwidth at render time), so a "fitting" line overflows the physical screen edge by ~1 col per emoji.

## Proposed fix (tested locally)

pi footers already support multiple lines (`render(width)` returns `string[]`), so overflow can wrap onto a continuation line instead of dropping content. Local patch in use since today, PTY-verified at 220/160/120/100/80 cols:

```ts
if (visibleWidth(left) > width) {
    segments = buildSegments(compactModel(model), theme);
    left = joinSegments(segments, theme);
}
// Wrap overflowing segments onto continuation lines instead of dropping
// them, so trailing statuses survive narrow (half-screen) terminals.
const lines: string[] = [];
let current: string[] = [];
// visibleWidth under-counts wide emoji by ~1 col each; a small margin
// absorbs that so a "fitting" line never physically clips at the edge.
const effectiveWidth = Math.max(20, width - 2);
for (const segment of segments) {
    if (current.length > 0 && visibleWidth(joinSegments([...current, segment], theme)) > effectiveWidth) {
        lines.push(current.length === 1 ? current[0] : joinSegments(current, theme));
        current = [segment];
    } else {
        current.push(segment);
    }
}
if (current.length > 0) lines.push(current.length === 1 ? current[0] : joinSegments(current, theme));
return lines.map((line, index) => truncateToWidth(index === 0 ? line : `  ${line}`, width, "…"));
```

Result at 100 cols (two lines, continuation indented):

```
✿ gentle-pi ⟡ tmp master ⟡ glm-5.3 · high ⟡ ctx ▲▲▲▲▲▲▲▲ 0% ⟡ $0.000 ⟡ 🧠 tmp · ready
  FFF indexed ⟡ 🔌 MCP: 7 servers enabled ⟡ LSP Inactive ⟡ ○ 🐴 ponytail: ⚡ FULL
```

The `-2` margin is a pragmatic band-aid; the deeper fix would be making `visibleWidth` count wide emoji correctly (a `String.codePointAt` + EAW table, or reusing pi-tui's width helper if one exists).

## Notes

- The `right`/`sessionName` early-return path is unchanged; wrapping only kicks in when the single line does not fit.
- Happy to open a PR with this (or an EAW-correct `visibleWidth`, or both) if the approach works for you — telling me which shape you prefer is enough.

AutanaSoft:
Part of #995.
Closes #715.

## Summary
- Reflow shell status metadata across one to three responsive footer lines.
- Preserve integration and omission indicators while truncating Unicode branch names safely.
- Add PTY and table-driven visible-width regression coverage.

## Changes
| File | Change |
| --- | --- |
| `lib/shell-bar.ts` | Reflow and safely truncate footer status content. |
| `tests/fixtures/shell-bar-pty.ts` | Exercise footer rendering through a PTY fixture. |
| `tests/shell-bar.test.ts` | Cover ANSI, Unicode, double-width, and narrow terminals. |
| `tests/gentle-shell.test.ts` | Cover shell integration behavior. |
| `docs/gentle-shell.md` | Document responsive footer behavior. |

## Validation
- [x] `pnpm test` — 2,373 passed, 11 skipped, 0 failed
- [x] `pnpm typecheck`
- [x] `pnpm run check:runtime-modules`
- [x] `node scripts/verify-package-files.mjs`
- [x] `pnpm run test:packed-package`
- [x] `git diff --check origin/main...HEAD`

## Delivery and rollback
This is the first of three ordered slices. Revert only this branch’s exclusive commits to restore the previous footer without removing structured RDD status or the compact sidebar.

## Summary by CodeRabbit

* **New Features**
  * Status-bar information now adapts across one to three lines based on terminal width.
  * Project, session, runtime, provider usage, cost, and integration details are arranged for improved readability.
  * Narrow terminals progressively simplify gauges and summarize overflow with indicators such as `+N more`.
  * Session names and other status text clip safely without exceeding the available width.
  * Sidebar status information now includes the active session.

* **Documentation**
  * Updated status-bar layout and subscription-usage guidance, including the compact five-cell gauge format.


danielgap:
Sharing mobile-width evidence here since this PR owns the responsive footer fix. The matrix in #715 stops at 80 cols; phone portrait goes further down and the drop gets worse.

Rendered through `renderShellBar` on current `main` (`b6188bef`, gentle-pi 3.3.0, pi 0.87.0), synthetic fixture:

| cols | rendered bar |
|---|---|
| 40-52 | `✿ gentle shell ⟡ gentleman main ±3` |
| 64 | `✿ gentle shell ⟡ gentleman main ±3 ⟡ claude-opus-4.6 · high` |
| 80 | `… ⟡ claude-opus-4.6 · high ⟡ ctx ▰▰▰▱▱▱▱▱ 42%` |
| 100+ | full bar including cost |

At 40-60 cols (typical phone portrait) model, context and cost are all gone, and the survivors are the brand and the compact location.

Two data points for the fix and its table-driven tests:

1. Priority inversion: the fullscreen header (`headerLeftStages`) already degrades in the right order, profile then effort then location give way before the model. The compact bar's pop loop does the opposite, so on the narrowest surfaces the most important session info is the first to go. Whatever shape lands here, model + ctx + cost should outlive brand + location.
2. #713 is the same ordering problem triggered by long Windows paths instead of narrow width. Worth one shared regression case if the fix lands here.

Heads-up: the branch reports CONFLICTING against `main` as of today, so it needs a rebase before review. Happy to verify the 40-60 col range against the rebased tip, that range is exactly the "very narrow terminals" case in the approved contracts on #995.

PR #1049, opened 2026-09-15
fix(changes): bound the Changes panel by files, not by mutations

### LINKED ISSUE #1043: bug(changes): session change capture stops after 256 write/edit records
### Before submitting

### Problem

After a long Pi session with gentle-pi (typical gentle-ai / SDD work with many `write`/`edit` calls, including managed subagents), the Changes panel shows:

```text
Warning: Session change capture limit reached; additional changes are not displayed.
```

Further agent edits stop appearing. The panel goes stale: already-listed files no longer update, and new files never appear.

This is the session-baseline capture from #765 (`lib/session-changes.ts`). Capacity is bounded on purpose (`MAX_RECORDS = 256`, `MAX_SESSION_BYTES = 4 MiB`). The 256 cap counts **mutation evidence records** (`toolCallId`s), not unique files. Each successful `write`/`edit` consumes one slot even when it only updates a file already in the panel. A long session therefore hits the warning well before 256 distinct files.

Related, not a duplicate: #765 (session-baseline diffs), #721 (Git refresh blocking the tool path). Duplicate search for the exact warning, `"capture limit"`, and `MAX_RECORDS` returned no matching issue.

### Steps to reproduce

1. Install Pi `0.85.1` with gentle-pi `2.7.0` (gentle-shell loaded).
2. Start an interactive session in any Git repository.
3. Have the agent perform more than 256 successful `write`/`edit` operations (repeats on the same files count). A long SDD/apply session or a synthetic loop of 257 unique-id records is enough. The unit test `capture capacity is explicit instead of growing without limit` already demonstrates the 257th record setting `notice`.
4. Open or watch the Changes widget.

Observed in a real long gentle-ai session on macOS; the synthetic 257-record path is the minimal reproduction.

Expected:
- The Changes panel keeps showing the **net** session-touched files (coalesced by path), or at least keeps updating files already being tracked.
- If a hard cap remains, the warning should say which limit fired (256 records vs 4 MiB) and that later mutations are dropped, without implying the session itself failed.

Actual:
- `SessionChanges.record()` sets `notice` to `Session change capture limit reached; additional changes are not displayed.` and returns `false` once `seen.size >= 256` or session snapshot JSON would exceed 4 MiB.
- Dropped records are not merged into existing file state, so the widget freezes at the last admitted mutations.

### gentle-pi version

2.7.0

### Pi version

0.85.1

### Operating system

macOS

```text
Warning: Session change capture limit reached; additional changes are not displayed.
```

Source (`lib/session-changes.ts`): `MAX_RECORDS = 256`, `MAX_SESSION_BYTES = 4 * 1024 * 1024`. The same notice string is assigned when either bound is hit. Tests assert 257 unique-path records yield 256 files and a `/limit reached/` notice; they do not cover many edits to few files, which is the long-session case.

L4XB:
Fixes #1043

## The problem

The Changes panel shows **files**. The cap counted **mutation evidence records**:

```ts
if (this.seen.size >= MAX_RECORDS || this.bytes + bytes > MAX_SESSION_BYTES) {
```

`seen` holds one `toolCallId` per admitted `write`/`edit`, so every repeat edit of a file already on screen spent one of the 256 slots. A long SDD session that touches a handful of files over and over reaches the warning while the panel lists maybe five entries, which is what the report describes.

The second half is what made it look frozen rather than merely capped. Once the bound fired, a later edit to a file the panel was **already showing** was dropped too, so that entry sat at its last admitted state. Refusing it bounded nothing: the `FileState` it would have updated is already held in `files`.

## The fix

`MAX_RECORDS` becomes `MAX_FILES`, checked against the number of distinct paths held, and only for a path that is not held yet:

```ts
let files = this.files.get(evidence.root);
const previous = files?.get(evidence.path);
if (!previous && this.trackedFiles >= MAX_FILES) {
  this.notice = FILE_LIMIT_NOTICE;
  return false;
}
```

An edit to a tracked file is now always admitted. What a long session can accumulate is still bounded, by the 4 MiB snapshot budget, which is unchanged.

The two bounds also stop sharing one sentence. The report asks for this and it is right: they have different remedies, and *"additional changes are not displayed"* told an operator neither which one filled up nor that the session itself was fine.

```
files:  Session change capture limit reached; 256 files are shown and later files
        are not tracked. Earlier files keep updating.
bytes:  Session change capture limit reached; the snapshot budget is full and
        later changes are not captured.
```

## Tests

`tests/session-changes.test.ts`, five cells added, and the existing 257-unique-path cell is untouched and still passes, since the bound is the same number for that shape.

| cell | what it stops |
| --- | --- |
| many edits to few files never reach the cap | the reported defect: 600 edits across 3 files, no notice, and a fourth file still admitted afterwards |
| a tracked file keeps updating after the cap fires | the staleness half, which a cap alone would leave in place |
| the notice names which bound fired | both bounds collapsing back into one sentence |
| a no-op edit to an unseen path does not spend a slot | `before === after` being charged for a file it never tracked |

Mutation results, 6 of 6 caught:

| mutation | result |
| --- | --- |
| the cap counts records again (the reported defect) | **caught** |
| a known file is refused once the cap is reached | **caught** |
| the file counter never advances | **caught** |
| the counter advances on every record, not only a new file | **caught** |
| both bounds share one notice again | **caught** |
| the byte budget is not enforced | **caught** |

The fourth is worth naming. Advancing the counter per record rather than per new file passes every other assertion in the file, and only surfaces as a *new* file being refused after 256 repeat edits, which is the original defect moved one line over. The first version of my own test did not catch it; the extra assertion in the many-edits cell is there because the mutation run said so.

`pnpm run typecheck`: 200 recorded diagnostics, no regressions. `tests/session-changes.test.ts` and `tests/session-changes-shell.test.ts`: 17 pass.

## What this does not change

The cap is still 256 and still hard. A session that genuinely touches more than 256 distinct files gets the same refusal, now with a notice that says so and does not imply the run failed. Coalescing without any bound was the other option in the report; I did not take it, because the panel's memory would then follow the session's file count with nothing holding it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Session changes are now limited by the number of distinct files rather than individual edits.
  - Repeated edits to already tracked files continue to be captured after the file limit is reached.
  - Capacity notices now clearly indicate whether the file-count or snapshot-size limit was reached.

- **Tests**
  - Added coverage for file-based limits, continued updates, available capacity, and no-op edits.


L4XB:
Merged `main` in to clear the conflict (merge commit, no rebase — the branch is already pushed).

Only `tests/session-changes.test.ts` conflicted, and only at the end of the file: #1096 and this branch each appended a new cell right after `test("capture capacity is explicit instead of growing without limit", ...)`, so the two blocks overlapped. Resolved by keeping #1096's `capture-limit notice is delivered once but stays on the model` exactly where `main` put it and re-appending this branch's four `#1043` cells after it — the file is now `main`'s version plus this branch's tests, with nothing dropped from either side.

`lib/session-changes.ts` merged without a conflict and holds both changes: `announced` / `takeNotice()` / `model.notice` from #1096, and the per-file bound with the two separate notices from here. `git diff main...HEAD --stat` is unchanged from before the merge — 2 files, 88 insertions(+), 5 deletions(-) — and the merged tree differs from `main` in exactly those two files.

`tests/session-changes.test.ts`, `tests/session-change-capture.test.ts` and `tests/shell-changes.test.ts`: 41/41 pass. `pnpm run typecheck` reports no regressions; `check:provider-contract`, `check:runtime-modules` and `verify-package-files.mjs` all pass. The remaining `pnpm test` failures on my machine are the `gentle-ai` native-binary cells, which fail identically on a clean `main` checkout here (no binary installed locally) — CI installs it.

PR #1050, opened 2026-09-15
fix(sdd): detect engram under every MCP tool-prefix mode

### LINKED ISSUE #1044: Engram tools are never detected when the MCP adapter prefixes tool names (hybrid preflight unreachable + SDD agents silently lose memory)
## Summary

`engramAvailable` is always `false` whenever the MCP tool-prefix mode is not `none` (the `pi-mcp-adapter` default is `server`). Two user-visible consequences, the second being silent data loss of agent memory.

## Environment

- gentle-pi 2.7.0
- pi 0.85.1
- pi-mcp-adapter 2.33.0
- Engram MCP server configured in `~/.pi/agent/mcp.json` with `"directTools": true`
- No `toolPrefix` key set on the server entry (so the adapter default `server` applies)

## Reproduction

1. Configure the Engram MCP server in `~/.pi/agent/mcp.json` **without** `"toolPrefix": "none"`.
2. Start Pi in any project and run `/gentle:sdd-preflight --edit`.
3. Observe the `SDD artifact store` prompt offers only one option.

Actual:

```
const artifactOptions = engramAvailable ? ["openspec", "engram", "hybrid"] : ["openspec"];
```

So `hybrid` and `engram` are not even offered, and any previously persisted `hybrid` value is silently normalized back to `openspec` (the on-disk `.pi/gentle-ai/sdd-preflight.json` is rewritten with `"engramAvailable": false`).

## Root cause

The probe compares one hardcoded tool name, but the adapter registers prefixed names:

```ts
// lib/sdd-preflight.ts:876-894 (duplicated at extensions/gentle-ai.ts:1883-1898)
function hasWritableEngramTool(pi: ExtensionAPI): boolean {
  ...
  return tools.some((tool) => {
    const name = /* tool.name */;
    return name === "mem_save" || name.endsWith(".mem_save");
  });
}
```

`pi-mcp-adapter` builds direct tool names as `formatToolName(toolName, serverName, prefix)`:

```ts
export function formatToolName(toolName, serverName, prefix) {
  const p = getServerPrefix(serverName, prefix);   // "server" | "short" | "mcp" -> non-empty; "none" -> ""
  const sanitized = toolName.replace(/\./g, "_");
  return p ? `${p}_${sanitized}` : sanitized;
}
// getServerPrefix: mode "none" -> "" ; "short" -> sanitized server name ; "mcp" -> `mcp__${server}` ; default "server" -> sanitized server name
```

With the default mode and a server named `engram`, the registered tool is `engram_mem_save`, which matches neither `mem_save` nor `*.mem_save`.

## Impact

**1. `engramAvailable` is wrong, so `hybrid` is unreachable** — including for a user who explicitly selected it. The artifact store silently degrades to `openspec`.

**2. Worse: every shipped SDD agent silently loses Engram.** The installed agent definitions declare:

```yaml
tools:
  - read
  - grep
  - find
  - write
  - edit
  - mem_search
  - mem_get_observation
  - mem_save
```

Those names do not exist under any prefix mode other than `none`, so the memory tools are dropped from the child's allowlist with no warning. Observed in practice: an `sdd-init` child reported *"No Engram MCP tools were injected"* and had to shell out to the `engram` CLI to persist its artifact, and an `sdd-research` child could not use `mem_search` / `mem_save` at all.

## Suggested fix

Resolve the Engram tools from the actual registry instead of hardcoding one name, for example:

- match a suffix that also tolerates the `_` separator: `/^(?:.*[._])?mem_save$/`; and
- preferably resolve the tool from the configured MCP server's tool inventory rather than from a name pattern.

The existing `endsWith(".mem_save")` suggests an earlier dot-separated naming scheme; it no longer matches the adapter's `_` separator.

## Workaround

Set the prefix to `none` on the Engram server entry in `~/.pi/agent/mcp.json`, then `/reload`:

```jsonc
"engram": {
  "command": "node",
  "args": ["..."],
  "directTools": true,
  "lifecycle": "lazy",
  "toolPrefix": "none"
}
```

Verified with a runtime inventory dump after the change: the tools register as `mem_save`, `mem_search`, `mem_get_observation`, etc., `engramAvailable` flips to `true`, `hybrid` becomes selectable, and the shipped agents' `mem_*` allowlist matches again.

## Evidence

Runtime inventory before/after the workaround, showing the registered names:

```
# default (toolPrefix: server)
engram_mem_save, engram_mem_search, engram_mem_get_observation, ...   # 18 tools
# toolPrefix: none
mem_save, mem_search, mem_get_observation, ...                        # 18 tools
```

L4XB:
Fixes #1044 (the preflight half; see **Scope** below)

## The problem

`pi-mcp-adapter` builds a direct tool name as `${prefix}_${tool}`, and only its `none` mode leaves the prefix empty. The probe compared one hardcoded spelling:

```ts
return name === "mem_save" || name.endsWith(".mem_save");
```

Under the adapter's **default** `server` mode the registered tool is `engram_mem_save`, which matches neither. So `engramAvailable` was `false` for everyone who had not set `"toolPrefix": "none"` by hand, `hybrid` and `engram` were never offered, and a previously persisted `hybrid` was rewritten back to `openspec` on disk without a word.

The dotted form the old code did match is an earlier naming scheme. It is kept.

## The fix

```ts
export function isEngramSaveToolName(name: string): boolean {
  return /^(?:.*[._])?mem_save$/.test(name);
}
```

| registered name | adapter mode |
| --- | --- |
| `mem_save` | `none` |
| `engram_mem_save` | `server` (the default) or `short` |
| `mcp__engram_mem_save` | `mcp` |
| `engram.mem_save` | the earlier dot-separated scheme |

Anchored at both ends deliberately. An unanchored suffix would accept `remem_save` and `mem_saved` and offer `hybrid` against a store that cannot persist anything, which is the same failure mode as the bug, pointed the other way.

The helper existed **twice**, in `lib/sdd-preflight.ts` and verbatim at `extensions/gentle-ai.ts:1883`, which your report notes and which is how one copy would have kept the old match after the other was fixed. The lib copy is exported now and the extension imports it, so the duplicate is gone.

## Tests

`tests/sdd-preflight.test.ts`, three cells.

| cell | what it stops |
| --- | --- |
| recognised under every adapter prefix mode | the reported defect, one assertion per mode |
| a name that merely contains `mem_save` is not the tool | the widening overshooting into `remem_save` / `mem_saved` |
| the registry read handles both shapes and its own absence | a host that returns `{ name }` records, one with no accessor, and one whose accessor throws |

Mutation results, 4 of 5 caught:

| mutation | result |
| --- | --- |
| the old bare/dotted match is back (the reported defect) | **caught** |
| the end anchor is lost, so `mem_saved` matches | **caught** |
| the separator is dropped, so `remem_save` matches | **caught** |
| only `_` counts, the dotted scheme is lost | **caught** |
| the `typeof getActiveTools !== "function"` guard is removed | **survived** |

The survivor is equivalent rather than a gap: without the guard the call throws a `TypeError` that the surrounding `catch` turns into the same `false`. I left the guard in place because it states the intent, rather than removing pre-existing defensive code in a bug fix.

`pnpm run typecheck`: 200 recorded diagnostics, no regressions. `tests/sdd-preflight.test.ts` and `tests/artifact-language.test.ts`: 46 pass.

For the record, `tests/gentle-agents.test.ts` has 5 failures on my machine. They are identical on a clean `main` with this branch stashed, so they are not this change, and I would rather say that than leave a green claim I did not get.

## Scope

This fixes the preflight probe, impact 1 in the report. Impact 2, the shipped SDD agents declaring bare `mem_save` / `mem_search` / `mem_get_observation` in their `tools:` allowlists, is a different mechanism: those names are load-bearing in `lib/sdd-research-capabilities.ts` for grant scoping, per-tool argument validation and the research persistence keys, in `extensions/gentle-agents.ts` for the write/edit settlement path, and in six `assets/agents/*.md` frontmatter blocks. Making that prefix-aware means a canonical-name mapping at the grant boundary rather than a wider pattern, and it lands in the same code the permission tests cover. I would rather propose that separately than attach it here. Happy to take it if you want it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the `/gentle:doctor` check to recognize writable Engram save tools across supported adapter naming formats.
  - Prevented the check from failing when tool information is unavailable or cannot be read.
  - Avoided incorrectly matching unrelated tools that merely contain similar text in their names.

- **Tests**
  - Added coverage for supported naming formats, tool representations, and unavailable registry scenarios.


L4XB:
On the outside-diff note about `hasWritableEngramTool` (`lib/sdd-preflight.ts`): right, the widened match took any server's `*_mem_save` for Engram. Fixed in e558205e.

`getActiveTools()` gives names only, and every `pi-mcp-adapter` tool shares the adapter's extension path, so provenance cannot tell one MCP server from another. The check now recognises Engram by its tool set instead. It takes a `mem_save` for Engram only when `mem_search` and `mem_get_observation` are active under the same prefix; those are the tools the SDD store reads its artifacts back with. That keeps every adapter prefix mode, and a server name the user chose, working.

New test: `another server's mem_save is not taken for Engram`. It covers `other_mem_save` alone, the set split across two prefixes, and a partial set, all rejected, plus the full set under a custom server name, accepted. It fails on the previous commit and when the tool-set requirement is removed.

The preflight test that stubbed a bare `mem_save` as "Engram available" now stubs the full set, so its premise still holds. `sdd-preflight.test.ts` passes 37 of 37, and `check-types.mjs` reports no new diagnostics.

PR #1161, opened 2026-09-17
fix(openspec): preserve internal markdown dividers in requirement content

Haiagari:
### Summary
Fixes an issue in `cleanRequirementContent` where internal markdown dividers/horizontal rules (`---`) within a requirement block were inadvertently stripped instead of the trailing boundary delimiter.

### Problem
In `lib/openspec-deltas.ts:31`, `cleanRequirementContent` uses:
```ts
return content.trimEnd().replace(/\n\s*---\s*$/m, "").trimEnd();
```
Because the regular expression uses the multiline flag `/m` without the global flag `/g` and is not anchored to the end of the entire string, `replace()` matches the **first** line that ends with `---`.

If a requirement body contains an internal separator—such as a markdown thematic break, table divider, or YAML code block—`replace()` removes that internal divider and leaves the trailing `---` delimiter at the bottom of the section, corrupting the requirement content.

### Solution
- Anchor the trailing delimiter replacement to the end of the full string:
  ```ts
  return content.trimEnd().replace(/(?:\r?\n\s*---\s*)+$/, "").trimEnd();
  ```
  This ensures any internal `---` lines remain untouched, and only the trailing delimiter separator at the end of the requirement block is stripped.
- Added a regression unit test in `tests/openspec-deltas.test.ts` to verify that internal markdown horizontal rules are preserved while trailing delimiters are stripped.

### Testing
- `node --experimental-strip-types --test tests/openspec-deltas.test.ts` passes with all 9 tests green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of trailing Markdown separators in requirement content, including consecutive separators, blank lines, and Windows-style line endings.
  - Internal Markdown horizontal rules are now preserved while only trailing separators are removed.

- **Tests**
  - Added coverage for preserving internal separators and removing trailing separators from requirement blocks.


barbatdev:
Thanks for this, the fix is small and the regression test makes the bug easy to see. One thing before we can consider approving it: could you open an issue for the bug and link it here? We need the issue approved first, then we can finish the review.

Haiagari:
Thanks @barbatdev! I have opened issue #1355 and linked it to this PR (Closes #1355). Ready whenever you can review.

PR #1243, opened 2026-09-19
fix(review): carry OpenCode session attribution into the in-process reviewer

### LINKED ISSUE #1242: Reviewer side-call drops OpenCode session attribution: every lens capture fails with 400 MissingSessionID
## What happened

Every lens capture whose routing model is `opencode/*` or `opencode-go/*` fails before any reviewer runs:

```text
400 {"type":"MissingSessionID",
     "message":"Request is missing x-opencode-session and cannot be routed efficiently."}
prepared_reviewers: 0, submitted_reviewers: 0
```

Two consecutive attempts failed identically — the second one after a fresh STATUS
re-offered the same binding. The outcome is `pi-host-relay-transport-failure` in
~440ms: no verdict, no approval burned, and no way to complete a review on such a
routing config.

## Root cause

Pi adds the OpenCode attribution headers **inside the main agent loop**:

`@earendil-works/pi-coding-agent` → `dist/core/provider-attribution.js`
- `getSessionHeaders(model, sessionId)` returns
  `{ "x-opencode-session": sessionId, "x-opencode-client": "pi" }` when the
  model provider is `opencode` / `opencode-go`, or the `baseUrl` host is
  `opencode.ai`.
- `mergeProviderAttributionHeaders(...)` merges it for the loop's own requests.

An extension that dispatches a **side call** bypasses that loop and must add the
same attribution itself. The in-process reviewer is exactly such a side call:
`lib/inprocess-reviewer.ts` builds its `SimpleStreamOptions` from only the
registry's `apiKey` and `headers`, so an OpenCode-routed model is called with no
session header at all.

A working reference for this same fix already exists in a sibling extension:
`pi-web-access/opencode-session-headers.ts` — *"Pi adds OpenCode attribution
headers inside the main agent loop. Extension side-calls dispatch directly, so
they must add the same session attribution."*

## Impact

- Any routing config that assigns an OpenCode model to a lens, refuter, or
  validator can never complete a review.
- Because the reviewer is a shared side-call path, this affects every repository
  the harness is installed in, not one project.

## Same class, wider surface

`getDefaultAttributionHeaders` (OpenRouter / NVIDIA NIM / Cloudflare) is bypassed
by the same side-call path. No failure has been reported for those, so only the
OpenCode leg is addressed here; worth a separate look.

## Fix

Thread the live session id from the extension context into the completion, and
add the attribution pair under the same condition pi's core uses, merged as a
default beneath the registry's auth headers. A missing session id adds nothing
and is never an error, and no other provider's options change.

PR follows and closes this issue.

IGabrielRC:
Closes #1242

## PR Type

- [x] Bug fix → `type:bug`

## Summary

- The in-process reviewer completion is an extension **side call**, so it
  bypasses pi's main agent loop — which is where OpenCode attribution headers are
  added. An OpenCode-routed reviewer model was therefore called with no
  `x-opencode-session`, and every lens capture failed with
  `400 MissingSessionID` before any reviewer ran.
- The reviewer now carries the live session id from the extension context into
  the completion and adds `{ x-opencode-session, x-opencode-client }` under the
  same condition pi's core uses (`opencode`, `opencode-go`, or a `baseUrl` host
  of `opencode.ai`), merged as a default beneath the registry's auth headers.
- A missing session id adds nothing and is never an error, and no other
  provider's options change.

## Changes Table

| File | Change |
| --- | --- |
| `lib/inprocess-reviewer.ts` | New exported `openCodeSessionAttributionHeaders(model, sessionId)` mirroring pi's `getSessionHeaders` condition; `InProcessReviewerRequest` gains `sessionId`; attribution merged beneath the registry auth headers |
| `lib/review-host-relay.ts` | `ReviewHostRelayRequest` gains `reviewerSessionId`, validated and mapped to the completion request's `sessionId` |
| `extensions/gentle-ai.ts` | Threads the live session id from the tool-handler context into both request builders (single-slot capture and reviewer group), reusing the existing safe extractor `reviewSessionManagerAndId` |
| `tests/inprocess-reviewer.test.ts` | 9 new tests: `opencode`, `opencode-go`, baseUrl-host, no-session-id for each of the three, non-OpenCode unchanged, registry headers win the merge, unparseable baseUrl never throws, helper parity with pi's condition |
| `tests/review-relay-transport-agent.test.ts` | The session id reaches the single-slot and group relay requests; an absent one forwards nothing |
| `tests/review-host-relay.test.ts` | Relay pass-through: the session id reaches the completion request, and an absent one forwards nothing |
| `odd/tasks/reviewer-opencode-session-headers.md` | Plan and verification record |

## Test Plan

- [x] `node --experimental-strip-types --test tests/inprocess-reviewer.test.ts` — 30 pass, 0 fail
- [x] `node --experimental-strip-types --test tests/review-relay-transport-agent.test.ts` — 14 pass, 0 fail
- [x] `node --experimental-strip-types --test tests/review-host-relay-routing.test.ts` — 31 pass, 0 fail
- [x] `pnpm run typecheck` — green, no regressions
- [ ] `pnpm test` — runs in CI's `verify` job on ubuntu; the two new relay tests live in a file whose harness spawns an extensionless fake binary, so that file is not part of the Windows contract
- [ ] End-to-end: a lens capture on an OpenCode-routed model completes instead of failing with `MissingSessionID` — needs a live OpenCode session after this lands

## Contributor Checklist

- [x] Conventional commit format (`fix(review): ...`)
- [x] Scope limited to the reviewer transport; Go admission, prompt materialization, schema, and submission untouched
- [ ] Issue approved (`status:approved` on #1242) — see the note below
- [ ] Exactly one `type:*` label on this PR — see the note below

## Notes for the maintainer

Two things required by the `branch-pr` skill could not be completed from the
contributing account (`IGabrielRC`, which has `pull` but not `push`/`triage` on
this repository, and no `AddLabelsToLabelable` permission):

1. `status:approved` on #1242, and exactly one `type:*` label on this PR.
2. The skill documents `PR Validation` jobs (`Check Issue Reference`,
   `Check Issue Has status:approved`, `Check PR Has type:* Label`), but they are
   not registered in this repository — the actions list shows only `CI`,
   `Publish to npm`, `Copilot`, and the two Windows workflows. So nothing
   automated blocks this PR; the labels are a house convention only.

Same class of gap, deliberately out of scope: `getDefaultAttributionHeaders`
(OpenRouter / NVIDIA NIM / Cloudflare) is bypassed by the same side-call path,
but no failure has been reported for those.

Also unrelated to this branch: `tests/gentle-shell.test.ts` carries a
pre-existing uncommitted change on the author's machine and is deliberately not
included here.

## Summary by CodeRabbit

- **New Features**
  - OpenCode-routed reviewer completions now retain live session attribution.
  - Session information is preserved across individual and grouped review captures.
  - Review capture requests remain compatible when no session identifier is provided.

- **Bug Fixes**
  - Prevented missing-session transport failures for in-process reviews using OpenCode-routed models.

- **Tests**
  - Added coverage for session forwarding, attribution headers, grouped captures, and existing provider behavior.

## Maintainer size exception

Approved for `size:exception`: 536 changed lines are justified because the production correction is narrowly scoped (77 lines), while 222 lines are behavior-focused regression tests across the three transport layers and 237 lines are the ODD task and verification record. Splitting the tests or evidence from the behavior would weaken reviewability and rollback confidence without creating an independent delivery boundary.


lordg95:
Independent reproduction and a negative result that may save review time.

**Same failure, different environment.** gentle-pi 3.3.0 facade, `pi-coding-agent` 0.86.0, `opencode-go` as the only configured provider, `deepseek-v4.1-flash` routed to `review-reliability`. The capture fails deterministically at the `pi` stage with `400 MissingSessionID`, `mutation_performed: false`, and the lineage left in `reviewing` — identical to what this PR describes and to the repro in #1260.

**The `sessionId` seam is confirmed to be enough on its own.** `pi-ai`'s `opencode-go.js` wraps its three APIs with `withOpenCodeSessionHeader`, and `withSessionHeader` turns `options.sessionId` into the header when it is not already present. I verified this against a local echo server: `completeSimple(model, context, { sessionId })` reaches the wire as `x-opencode-session`, and an absent `sessionId` is omitted rather than sent empty. So carrying the live session id is what makes the completion routable; adding the headers under a provider/host test is belt-and-braces on top of that, not a requirement.

**No configuration can fix this** (tried before writing anything, in an isolated `PI_CODING_AGENT_DIR`): provider-level `headers` through `pi.registerProvider`, provider-level `headers` in `models.json`, `modelOverrides: { "*": { headers } }`, and full per-model definitions each carrying `headers` — all four left **0 of 27** `opencode-go` models with headers. Config headers resolve into the auth resolution and are applied in pi's request layer, which is exactly the layer this completion bypasses. The `before_provider_headers` extension hook lives inside that same layer, so it cannot help either.

**A gotcha worth knowing when testing this.** Before the completion is ever reached, the routing config has to carry an entry for the lens agent itself (`review-risk`, `review-resilience`, `review-readability`, `review-reliability` in `~/.pi/gentle-ai/models.json`). Without one the capture is refused typed with `reviewer-config-invalid` naming the key, which looks like a different bug.

I have a branch implementing the same fix through the `sessionId` seam alone (`fix/review-session-id`, 4 files, +55/-2, tests 22/22). Standing down in favour of this PR — no competing PR from me. Happy to hand over the echo-server probe or the test cases if they are useful here.

Alan-TheGentleman (approved):
Approved. The session attribution is threaded through single and grouped reviewer captures, the OpenCode routing condition and header precedence match Pi core, and the regression coverage exercises every transport layer. Local verification and all CI jobs are green.

PR #1252, opened 2026-09-20
fix(shell): show effective repository profile source

### LINKED ISSUE #1176: bug(shell): Status shows the global profile instead of the repository-effective profile
### Before submitting

### Problem

The fullscreen Status sidebar displays the globally active agent-model profile from `profiles.json.active`, even when the current repository has a valid clone-local pin or repository declaration selecting a different profile.

Repository pins replace global subagent routing at launch. Therefore, the displayed profile can disagree with the profile actually governing subagent launches in the current repository. This makes the Status view misleading when per-repository isolation is being used.

The current data flow is:

- `extensions/gentle-shell.ts:createActiveProfileReader()` reads only the global profiles store and returns `file.active`.
- It does not resolve the current repository through `resolveProfilePin()`.
- `lib/shell-bar.ts` renders that value as `Profile <name>` in the fullscreen Status sidebar.

This appears to be an interaction between #1000, which added the active-profile indicator, and #1074, which later added per-repository profile pins.

### Steps to reproduce

1. Create two agent-model profiles, for example `global-default` and `project-specific`.
2. Apply `global-default` globally through `/gentle:profiles`.
3. Inside a Git repository, select `project-specific` and press `p` to create the clone-local profile pin. Alternatively, press `P` to create the repository declaration.
4. Confirm through `/gentle:profiles` that the repository resolves `project-specific` as its pinned profile.
5. Open the fullscreen Status sidebar.
6. Observe the `Profile` field.
7. Launch a subagent from the repository and compare its resolved routing with the profile displayed by Status.

Expected:

The Status sidebar should use its existing `Profile` field to display the effective profile governing subagent launches in the current repository.

When the repository has no valid profile pin:

```text
Profile  global-default
```

When a clone-local pin or repository declaration wins:

```text
Profile  project-specific (pinned)
```

The `(pinned)` suffix should appear only when the effective profile comes from a valid repository pin. No additional field for the global profile is needed.

Actual:

The Status sidebar always displays `profiles.json.active`. In a pinned repository, it therefore shows:

```text
Profile  global-default
```

while subagent launches resolve `project-specific`.

The compact bottom bar does not currently display a profile, so this mismatch affects the fullscreen Status sidebar.

### gentle-pi version

```text
3.2.0 source checkout, commit 2b579c80824e
```

### Pi version

```text
0.85.1
```

### Operating system

```text
Linux
```

```text
No runtime error is emitted.

The mismatch is visible in the resolved state:

Global profiles store:
  active = global-default

Repository profile resolution:
  winner = project-specific
  source = local or repo

Fullscreen Status sidebar:
  Profile global-default
```

Related issues:

- #1000 added the active-profile indicator to the Status view.
- #1074 added per-repository profile pins.
- #995 covers unified responsive Status presentation but not effective profile resolution.
- #1064 proposes session-bound profiles and broader effective-profile source visibility, but does not cover this existing repository-pin mismatch.

AutanaSoft:
Closes #1176

## Summary

- Show the repository-effective profile in the fullscreen Status card and live header.
- Preserve the winning profile source as `(local)` or `(repo)`.
- Refresh profile state outside render paths when profile or pin files change.
- Keep the compact bottom bar unchanged.

## Problem

The fullscreen shell surfaces previously displayed the globally active profile even when repository
routing resolved a clone-local pin or repository declaration.

The display also collapsed both repository scopes into a generic `(pinned)` marker, despite
`resolveProfilePin()` already identifying the winning source.

## Behavior

The effective profile now follows the existing routing precedence:

```text
clone-local pin → repository declaration → global active profile
```

The fullscreen surfaces render:

```text
Profile global-default
Profile project-specific (local)
Profile project-specific (repo)
```

Removing a local pin may reveal an existing repository declaration. The global profile becomes
effective only when neither pin layer resolves.

## Changes

| File | Change |
| --- | --- |
| `extensions/gentle-shell.ts` | Cache and refresh the effective profile outside render paths while preserving its source. |
| `lib/shell-bar.ts` | Represent profile state as `global`, `local`, or `repo` and format the visible suffix. |
| `tests/gentle-shell.test.ts` | Cover resolution, refreshes, atomic replacements, source transitions, and render-path isolation. |
| `tests/shell-bar.test.ts` | Cover profile formatting and compact-bar stability. |
| `docs/gentle-shell.md` | Document fullscreen profile-source behavior. |
| `docs/readme-reference.md` | Document precedence, suffixes, and fallback behavior. |

## Test plan

- [x] Focused shell suites: 160 passed, 0 failed.
- [x] Typecheck passed with no new diagnostics.
- [x] Full unit suite passed using the documented split for the unavailable package-local native runtime.
- [x] Relay-routing suite passed: 31 passed, 0 failed.
- [x] Remaining unit suite passed: 2816 passed, 0 failed, 47 skipped.
- [x] `git diff --check` passed.
- [x] Verified that repeated Status/header renders perform no profile filesystem or Git resolution.
- [x] Verified same-profile transitions between local and repository sources.
- [x] Documentation updated.

## Size exception request

This PR changes 720 lines:

| Category | Added | Removed | Total |
| --- | ---: | ---: | ---: |
| Production code | 254 | 33 | 287 |
| Tests | 327 | 19 | 346 |
| Documentation and ODD evidence | 87 | 0 | 87 |

Most of the diff is behavior-focused regression coverage. The implementation, watcher lifecycle,
render-path isolation, and tests form one cohesive correction for #1176. Splitting them would
separate the behavior from the tests that establish its safety.

Please approve `size:exception` if this review boundary is acceptable.

## Checklist

- [x] Linked an approved issue.
- [x] Uses a Conventional Commit message.
- [x] Added focused regression coverage.
- [x] Updated behavior documentation.
- [x] No shell scripts were modified.
- [x] No `Co-Authored-By` trailers were added.


jals1212:
Tested this on macOS 27.0 with Pi 0.86.1, using a local merge of this branch (`cfbdf1be`) on top of current `main` (3.6.0). The merge was clean.

**Manual check: works as expected** ✅
I used a real repository with a clone-local pin (`profile-pin.json` → `openai`), while the globally active profile was `current`. I launched it with `gentle-shell --link --package-root <checkout>`.
1. Status and the header showed `Profile openai (local)`. The published 3.5.1 shows `Profile current`.
2. Removing the pin with `p` in `/gentle:profiles` switched the display to `current` with no suffix, without restarting.
3. Pinning again with `p` switched it back to `openai (local)`.

**Automated checks**
- `pnpm run typecheck`: no regressions. `check:provider-contract` and `test:harness` pass.
- `pnpm test`: 3351 pass and 2 fail, and both failures are tests added in this PR. They are flaky here and always pass when run on their own with `--test-name-pattern`:
  - `fullscreen Status digest follows effective pin source changes without restarting the shell`: `timed out waiting for profile "name":"other","source":"repo"`. It failed 7/7 times when the whole file ran.
  - `fullscreen Status rebinds parent watchers when a profile store appears after startup`: `timed out waiting for profile "name":"team"`. It failed about half the time.

  These tests wait on real `fs.watch`/FSEvents plus the 100 ms debounce, and give it only 400–600 ms in total. On macOS, events can arrive late, or get lost when a write happens right after the watch is set up. Two options: poll with a longer deadline, or cover the refresh logic through the injectable `watch` seam instead of the real watcher.

**Non-blocking observations**
- Each refresh calls `resolveWorktree(cwd, cwd)` through `effectiveProfileWatchDirectories`, and `resolveProfilePin` resolves it again. That means synchronous `git rev-parse` calls on the main thread on every watch event and branch change. Caching the identity for each cwd would avoid them.
- When `<commonDir>/gentle-ai` or `.pi/` doesn't exist, which is the common case, the watcher falls back to `.git` or the repository root. Ordinary git or editor activity then keeps triggering refreshes, each with the git calls above.

Thanks for the fix, it resolves #1176 for my setup.

